Issue #4, opened 2019-12-16
markdown generation


gumaerc:
As a user of this command, I would like to be able to pass it source and destination directories and have it output Hugo compatible markdown based on the data in the source directory to the destination.

#### Acceptance Criteria:

- [ ] If the command is run against a directory with course folders inside of it, each containing a file whose name contains `_master.json`, the command should output Hugo compatible markdown for each of these courses to the destination directory

PR #5, opened 2019-12-16
Basic markdown creation

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/4

#### What's this PR do?
Adds classes for generating markdown data from the master JSON file as well as some helpers for writing that data to disk

Follow the installation and usage instructions in the readme.  Note that you will need to switch to the `cg/process-json` branch before installing.


lib/markdown_generators.js
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env node
+
+const yaml = require("js-yaml");
+const markdown = require("markdown-builder");
```
noisecapella: It looks like this library's repo is archived by the owner. Is there an alternative we can use?

lib/markdown_generators.js
```diff
@@ -0,0 +1,146 @@
...
+const turndown_service = new TurndownService();
+turndown_service.use(tables);
+const helpers = require("./helpers");
+
+class markdown_generators {
+    static generate_course_home_front_matter(course_data) {
+        /*
+        Generate the front matter metadata for the course home page given course_data JSON
+        */
+        let front_matter = {};
```
noisecapella: We should assign the values in the initial object, to the extent possible:

    const front_matter = {
      title: "Course Home",
      course_title: course_data['title'],
      course_image_url: helpers.get_course_image_url(course_data),
      ...
    }

lib/markdown_generators.js
```diff
@@ -0,0 +1,146 @@
...
+        front_matter["title"] = "Course Home";
+        front_matter["course_title"] = course_data["title"];
+        front_matter["course_image_url"] = helpers.get_course_image_url(course_data);
+        front_matter["course_description"] = course_data["description"];
+        front_matter["course_info"] = {};
+        front_matter["course_info"]["instructors"] = [];
```
noisecapella: This can be

    instructors: course_data["instructors"].map(instructor => `Prof. ${instructor['first_name']} ${instructor['last_name']}`)

lib/markdown_generators.js
```diff
@@ -0,0 +1,146 @@
...
+        course_data["instructors"].forEach(instructor => {
+            front_matter["course_info"]["instructors"].push(
+                "Prof. " + instructor["first_name"] + " " + instructor["last_name"]
+            );
+        });
+        front_matter["course_info"]["department"] = title_case.titleCase(
+            course_data["url"].split("/")[2].replace(/-/g, " ")
+        );
+        front_matter["course_info"]["topics"] = [];
```
noisecapella: This can be

    topics: course_data["course_collections"].map(makeTopic)`

And `makeTopic` the function in the next line, with `return topic` in place of the `front_matter[...].push(...)`

lib/markdown_generators.js
```diff
@@ -0,0 +1,146 @@
...
+const yaml = require("js-yaml");
+const markdown = require("markdown-builder");
+const title_case = require("title-case");
+const TurndownService = require("turndown");
+const tables = require("turndown-plugin-gfm").tables;
+const turndown_service = new TurndownService();
+turndown_service.use(tables);
+const helpers = require("./helpers");
+
+class markdown_generators {
```
noisecapella: This is redundant for the file. We can export the functions below directly instead, for example

    module.exports = {
      generate_course_home_front_matter,
      generate_course_section_front_matter
      ...
    }

lib/markdown_generators.js
gumaerc: This is a simple string builder type thing that has convenience functions for generating markdown snippets.  I haven't found anything exactly like it, unfortunately.  I could just use string literals to build the markdown I need, rather than depend on the library but it could get tedious depending on what special casing needs to be done once we get in to the thick of looking at individual courses.

src/lib/file_operations.js
```diff
@@ -0,0 +1,94 @@
...
+)
+let totalDirectories = 0
+let directoriesScanned = 0
+let filesProcessed = 0
+
+const scanCourses = (source, destination) => {
+  /*
+    This function scans the source directory for course folders
+  */
+  fs.readdir(source, (err, directories) => {
```
noisecapella: Can we use `readdirSync` here and below? Alternatively we can change these to async functions and use util.promisify, but I'm not sure that would be necessary here.

src/lib/file_operations.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+  })
+}
+
+const writeMarkdownFiles = (courseId, markdownData, destination) => {
+  /*
+    For a given course identifier string and array of objects with properties 
+    name and data, write Hugo markdown files
+    */
+  if (destination && fs.lstatSync(destination).isDirectory()) {
```
noisecapella: `destination` should always exist since we specify it as a required argument. We should check if `destination` is a directory closer to the entry point of the script and error at that point if something went wrong. At this point it is checked for every master JSON file which isn't necessary.

src/lib/file_operations.js
```diff
@@ -0,0 +1,94 @@
...
+let filesProcessed = 0
+
+const scanCourses = (source, destination) => {
+  /*
+    This function scans the source directory for course folders
+  */
+  fs.readdir(source, (err, directories) => {
+    // Count the total amount of directories and start the progress bar
+    directories.forEach(directory => {
+      const coursePath = source + directory
```
noisecapella: We should use [path.join](https://nodejs.org/docs/latest-v10.x/api/path.html#path_path_join_paths) instead of string concatenation for concatenating paths together.

src/lib/helpers.js
```diff
@@ -0,0 +1,41 @@
...
+  }${courseNameParts[courseNameParts.length - 2].charAt(0)}${courseNameParts[
+    courseNameParts.length - 1
+  ].slice(2)}.jpg`
+  courseData["course_files"].forEach(media => {
+    if (media["parent_uid"] === courseData["uid"]) {
+      const fileLocationParts = media["file_location"].split("/")
+      const jsonFile = fileLocationParts[fileLocationParts.length - 1]
+      const imageFile = `${media["uid"]}_${imageName}`
+      if (jsonFile === imageFile) {
+        return media["file_location"]
```
noisecapella: This will end the `forEach` function here, not `getCourseImageUrl`. Note that you can also use `for ... of` here to avoid this problem:

    for (const media of courseData["course_files]) {
      ...
    }

src/lib/helpers.js
```diff
@@ -0,0 +1,41 @@
...
+      }
+    }
+  })
+  return "images/course_image.jpg"
```
noisecapella: Is this true for each OCW file?

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,182 @@
...
+      Generate the front matter metadata for the course home page given course_data JSON
+      */
+
+  let courseNumber = courseData["sort_as"]
+  if (courseData["extra_course_number"]) {
+    if (courseData["extra_course_number"]["sort_as_col"]) {
+      courseNumber += ` / ${courseData["extra_course_number"]["sort_as_col"]}`
+    }
+  }
+  const makeTopic = feature => {
```
noisecapella: We should move this to the top level of the file, otherwise this will be recreated for each function call for `generateCourseHomeFrontMatter`. We should only define functions inside other functions if there is some context we need to access here.

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,182 @@
...
+
+const generateCourseFeatures = courseData => {
+  /*
+      Generate markdown for the "Course Features" section of the home page
+      */
+  let courseFeaturesMarkdown = markdown.headers.hX(5, "Course Features")
+  const courseFeatures = []
+  courseData["course_features"].forEach(courseFeature => {
+    const url = courseFeature["ocw_feature_url"]
+      .replace("/index.htm", "/")
```
noisecapella: If there is an `index.html` file this would leave a stray `l`.

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,182 @@
...
+      .split("/")[-1]
```
noisecapella: Negative indexing doesn't work in JS like it does in Python. You will need to do something like `urlPieces[urlPieces.length - 1]`, but maybe there is a better way to do something like this.

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+  })}---\n`
+}
+
+const generateCourseFeatures = courseData => {
+  /*
+      Generate markdown for the "Course Features" section of the home page
+      */
+  let courseFeaturesMarkdown = markdown.headers.hX(5, "Course Features")
+  const courseFeatures = []
```
noisecapella: Can we use `.map(...)` below to change this to something like:

    const courseFeatures = courseData["course_features"].map(courseFeature => 
      courseFeature["ocw_feature_url"].replace("/index.htm", "/").split("/")[-1]
    )

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,182 @@
...
+    const subfeature = courseCollection["ocw_subfeature"]
+    const specialty = courseCollection["ocw_specialty"]
+    let collection = feature
+    if (subfeature) {
+      collection += ` > ${subfeature}`
+    }
+    if (specialty) {
+      collection += ` > ${specialty}`
+    }
+    courseCollections.push(markdown.misc.link(collection, "#"))
```
noisecapella: Can we change this function from a `forEach` to `map` since it's populating `courseCollections`?

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,182 @@
...
+
+const generateCourseSectionMarkdown = (page, courseData) => {
+  /*
+      Generate markdown a given course section page
+      */
+  let htmlStr = page["text"]
+  courseData["course_pages"].forEach(coursePage => {
+    const placeholder = `./resolveuid/${coursePage["uid"]}`
+    const pageName = coursePage["short_url"].replace(".html", "")
+    if (htmlStr.indexOf(placeholder) > -1) {
```
noisecapella: We can do `if (htmlStr.includes(placeholder)) {`

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,182 @@
...
+}
+
+const generateCourseSectionMarkdown = (page, courseData) => {
+  /*
+      Generate markdown a given course section page
+      */
+  let htmlStr = page["text"]
+  courseData["course_pages"].forEach(coursePage => {
+    const placeholder = `./resolveuid/${coursePage["uid"]}`
+    const pageName = coursePage["short_url"].replace(".html", "")
```
noisecapella: What is the `replace` doing here?

src/lib/file_operations.js
noisecapella: Or [fsPromises](https://nodejs.org/api/fs.html#fs_fspromises_readdir_path_options) in place of `util.promisify`

src/lib/markdown_generators.js
gumaerc: This is an artifact from the code I ported from ocw-data-parser.  The function this is used in there looks at all the pages, including the index, which has a ".html" in the short_url.  It's not necessary here so I'll remove it.

src/lib/markdown_generators.js
gumaerc: This is another artifact from ocw-data-parser code.  I will change it to use regex and optionally include "/index.html" though, as I'm not sure that .htm is the only extension used here.

src/lib/helpers.js
gumaerc: It's not.  I put this in there as a fallback default because I had a brain fart about returning from forEach.  I altered the code so that this is just set on a variable as the default, then the forEach loop will set that variable if the file is actually found in the course_files collection.

src/bin/index.js
```diff
@@ -18,30 +21,13 @@ const options = yargs
     type:         "string",
     demandOption: true
   }).argv
+// Ensure that there is a trailing slash on the source and destination paths
```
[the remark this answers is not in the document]
gumaerc: Not needed, I removed it after you reminded me about path.join and just used that instead.

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,184 @@
...
+      let courseSectionMarkdown = generateCourseSectionFrontMatter(
+        page["title"],
+        menuIndex
+      )
+      courseSectionMarkdown += generateCourseSectionMarkdown(page, courseData)
+      const sectionData = {
+        name: `sections/${pageName}.md`,
+        data: courseSectionMarkdown
+      }
+      markdownData.push(sectionData)
```
noisecapella: We can make this a map and use the index provided as a second parameter for `menuIndex`:

    const coursePagesWithText = courseData["course_pages"].filter(page => page["text"])
    coursePagesWithText.map((page, menuIndex) => {
      const courseSectionMarkdown = generateCourseSectionFrontMatter(
        page["title"],
        (menuIndex + 1) * 10
      )
      courseSectionMarkdown += generateCourseSectionMarkdown(page, courseData)
      return {
        name: `sections/${pageName}.md`,
        data: courseSectionMarkdown
      }
    })

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,184 @@
...
+}
+
+const generateCourseFeatures = courseData => {
+  /*
+      Generate markdown for the "Course Features" section of the home page
+      */
+  const courseFeaturesHeader = markdown.headers.hX(5, "Course Features")
+  const courseFeatures = courseData["course_features"].map(courseFeature => {
+    const urlParts = courseFeature["ocw_feature_url"]
+      .replace(/\/index.htm?l/, "/")
```
noisecapella: I know you said this came from ocw-data-parser but can you point me to where this came from? I am curious about the context here.

src/lib/markdown_generators.js
```diff
@@ -0,0 +1,184 @@
...
+  courseData["course_pages"].forEach(coursePage => {
+    const placeholder = `./resolveuid/${coursePage["uid"]}`
+    if (htmlStr.includes(placeholder)) {
+      htmlStr = htmlStr.replace(
+        placeholder,
+        `/sections/${coursePage["short_url"]}`
+      )
+    }
+  })
+  courseData["course_files"].forEach(media => {
```
noisecapella: We should cut this part until we need it later, when we have resolved the TODO question below

src/bin/index.js
```diff
@@ -19,29 +21,9 @@ const options = yargs
...
-              processJson(courseData)
-            }
-          })
-        })
-      }
-    })
-  })
+  scanCourses(options.source, options.destination)
 } else {
   console.log("Invalid source directory")
```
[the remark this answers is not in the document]
gumaerc: I had removed this, but it looks like it got shoved back by accident in when I rebased

src/lib/markdown_generators.js
gumaerc: https://github.com/mitodl/ocw-data-parser/blob/master/ocw_data_parser/static_html_generator.py#L146

src/lib/helpers.js
```diff
@@ -0,0 +1,40 @@
...
+  if (lastChar !== "/") {
+    return `${path}/`
+  } else return path
+}
+
+const getCourseImageUrl = courseData => {
+  /*
+    Constructs the course image filename using parts of the course short_url
+    */
+  const courseNameParts = courseData["short_url"].split("-")
```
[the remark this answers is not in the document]
gumaerc: I don't think it would be particularly useful here, and would probably add more lines of code but this is good to know.

src/lib/helpers.js
```diff
@@ -0,0 +1,40 @@
...
+  const imageName = `${courseNameParts[0]}-${
+    courseNameParts[1]
+  }${courseNameParts[courseNameParts.length - 2].charAt(0)}${courseNameParts[
```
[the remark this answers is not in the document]
gumaerc: Per our convo in slack:

for example:
`courseData["short_url"] == "10-520-molecular-aspects-of-chemical-engineering-fall-2004"`
`imageFile == "17cc828df8e9a5986ec838e785ef61b8_10-520f04.jpg"`

so the course image file structure is plone_uid followed by an underscore, then the "course number," then 1 letter and 2 numbers representing the term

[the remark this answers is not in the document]
gumaerc: https://github.com/mitodl/ocw-data-parser/blob/master/ocw_data_parser/static_html_generator.py#L120

src/lib/file_operations.js
```diff
@@ -0,0 +1,94 @@
...
+      if (fs.lstatSync(coursePath).isDirectory()) {
+        // If the item is indeed a directory, read all files in it
+        scanCourse(coursePath, destination)
+      }
+    })
+  })
+}
+
+const scanCourse = (coursePath, destination) => {
```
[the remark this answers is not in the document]
gumaerc: I'm not certain I agree with this one.  We may need some clarification from OCW, as there may be a reason master json files aren't generated for certain courses.  I think error logging should be a separate PR that sets up winston for logging these types of things to a file.

src/lib/file_operations.js
```diff
@@ -0,0 +1,94 @@
...
+      }
+    })
+    console.log(`Scanning ${totalDirectories} subdirectories under ${source}`)
+    progressBar.start(totalDirectories, directoriesScanned)
+    // Iterate all subdirectories under source
+    directories.forEach(directory => {
+      const coursePath = source + directory
+      if (fs.lstatSync(coursePath).isDirectory()) {
+        // If the item is indeed a directory, read all files in it
+        scanCourse(coursePath, destination)
```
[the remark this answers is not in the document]
gumaerc: You mean if no courses at all are found in the source folder?

src/lib/file_operations.js
gumaerc: Punting on this, per our discussion about the memory leak issue with turndown.

src/lib/helpers.js
```diff
@@ -10,18 +10,17 @@ const getCourseImageUrl = courseData => {
   }${courseNameParts[courseNameParts.length - 2].charAt(0)}${courseNameParts[
     courseNameParts.length - 1
   ].slice(2)}.jpg`
-  let courseImageUrl = "images/course_image.jpg"
-  courseData["course_files"].forEach(media => {
+  const courseImageMedia = courseData["course_files"].filter(media => {
```
noisecapella: Actually, you can use `find` here if you want to get the first item which matches a condition

src/lib/markdown_generators.js
```diff
@@ -33,22 +33,23 @@ const generateMarkdownFromJson = courseData => {
...
-      const pageName = page["short_url"]
-      let courseSectionMarkdown = generateCourseSectionFrontMatter(
-        page["title"],
-        menuIndex
-      )
-      courseSectionMarkdown += generateCourseSectionMarkdown(page, courseData)
-      const sectionData = {
-        name: `sections/${pageName}.md`,
-        data: courseSectionMarkdown
+  markdownData.concat(
```
noisecapella: Should be `return markdownData.concat(...`

noisecapella: Oh I missed the `return markdownData` below. However `concat` will not mutate the array so you need to `return markdownData.concat` here or `markDownData = markdownData.concat(...)` or `.push` the items one by one

noisecapella: Another way is to use the spread operator:

    return [
      {
        name: "_index.md",
        data: courseHomeMarkdown
      },
      ...coursePagesWithText.map(...)
    ]

PR #9, opened 2019-12-18
Remove babel

noisecapella:
#### Pre-Flight checklist

- [x] Testing

None

#### What's this PR do?
Removes babel. Since this is a command line script running on node, almost every ES2015 feature should already be supported. Since this a small project we will use `require` instead of `import` to avoid having users run node with the experimental module flag (more info [here](https://thecodebarbarian.com/nodejs-12-imports)).

Tests should pass, and the command should run without errors and generating expected output on sample courses

### DIFF
--- a/scripts/js_test.sh
+++ b/scripts/js_test.sh
@@ -15,7 +15,7 @@ else
 fi
 
 export FILE_PATTERN=${1:-'"src/**/*_test.js"'}
-CMD_ARGS="--require @babel/register $FILE_PATTERN"
+CMD_ARGS="$FILE_PATTERN"
 
 # Second argument (if specified) should be a string that will match specific test case descriptions
 #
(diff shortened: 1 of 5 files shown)


scripts/js_test.sh
```diff
@@ -15,7 +15,7 @@ else
 fi
 
 export FILE_PATTERN=${1:-'"src/**/*_test.js"'}
-CMD_ARGS="--require @babel/register $FILE_PATTERN"
+CMD_ARGS="$FILE_PATTERN"
```
aliceriot: are we going to bother with coverage and so on in this project? if no we can delete this whole file and call mocha directly from the `package.json`

noisecapella: I think we should have coverage and @gumaerc is working on it now. It would be useful because I think the complicated business logic will need significant testing.

Issue #10, opened 2019-12-19
Handle HTML tables in course sections


gumaerc:
The various course section pages brought into the master JSON file contain some HTML tables. Hugo supports rendering HTML tables from markdown tables, however there are various HTML tags inside the table cells that are problematic to represent in markdown:

- Paragraphs
- Ordered / unordered lists
- Divs
- etc.

Tables in markdown are represented as such:
```
| Tables   |      Are      |  Cool |
|----------|:-------------:|------:|
| col 1 is |  left-aligned | $1600 |
| col 2 is |    centered   |   $12 |
| col 3 is | right-aligned |    $1 |
```

Each line is interpreted as a row, so all content for a table cell needs to be on one row.  The turndown tables plugin does not do anything special when converting HTML tags inside tables, and with elements like paragraphs (for example) it will insert two line breaks with the content in between.  This breaks the markdown table representation and Hugo renders it as a messy block of text with pipes everywhere.

PR #13, opened 2019-12-30
Table processing

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/10

#### What's this PR do?
Adds a turndown rule to process content in raw HTML tables brought in from OCW course sections so they are rendered properly by Hugo.

Run any number of courses through this tool, then take the markdown output and load it into hugo-course-publisher.  The assignments or readings sections usually have tables, and they should now display properly.

image

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -44,8 +44,8 @@ turndownService.addRule("table", {
      */
     content = content
       .substring(content.indexOf("|"), content.lastIndexOf("|"))
-      .replace(/\r?\n|\r/g, "<br>")
-      .replace(/\|<br>\|/g, "|\n|")
+      .replace(/\r?\n|\r/g, "{{< br >}}")
+      .replace(/\|{{< br >}}\|/g, "|\n|")
       .replace(/REPLACETHISWITHAPIPE/g, "&#124;")
     // Only do this if header lines are found
     if (content.match(/(?<=\| \*\*)(.*?)(?=\*\* \|\r?\n|\r)/g)) {
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -23,10 +23,10 @@ turndownService.addRule("table", {
     // then reintroduce them between rows
     content = content
       .substring(content.indexOf("|"), content.lastIndexOf("|"))
-      .replace(/\r?\n|\r/g, "")
-      .replace(/\|\|/g, "|\r\n|")
+      .replace(/\r?\n|\r/g, "{{<br>}}")
+      .replace(/\|{{<br>}}\|/g, "|\r\n|")
     // Only do this if header lines are found
-    if (content.match(/(?<=\| \*\*)(.*?)(?=\*\* \|\r?\n)/g)) {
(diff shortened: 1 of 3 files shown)


src/markdown_generators.js
```diff
@@ -4,35 +4,66 @@ const yaml = require("js-yaml")
...
+     * from earlier with the HTML character entity for a pipe
+     * reintroduce them between rows, replacing our pipe marker
+     */
+    content = content
+      .substring(content.indexOf("|"), content.lastIndexOf("|"))
+      .replace(/\r?\n|\r/g, "{{< br >}}")
+      .replace(/\|{{< br >}}\|/g, "|\n|")
+      .replace(/REPLACETHISWITHAPIPE/g, "&#124;")
+    // Only do this if header lines are found
+    if (content.match(/(?<=\| \*\*)(.*?)(?=\*\* \|\r?\n|\r)/g)) {
```
aliceriot: some of these longer / more complicated regexes it might be a good idea to pull out as a separate viable and just writing a little comment describing what they're supposed to match

src/markdown_generators.js
```diff
@@ -4,35 +4,66 @@ const yaml = require("js-yaml")
...
+     */
+    for (let i = 0; i < node.rows.length; i++) {
+      const cells = node.rows[i].cells
+      for (let j = 0; j < cells.length; j++) {
+        cells[j].innerHTML = cells[j].innerHTML.replace(
+          /\|/g,
+          REPLACETHISWITHAPIPE
+        )
+      }
+    }
```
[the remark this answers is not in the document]
gumaerc: I avoided doing it this way specifically because the "node" object passed into this function is a native DOM HTMLElement.  The code actually breaks if I put in what you have above, because forEach isn't a function on `.rows` or `.cells`.

src/markdown_generators.js
```diff
@@ -82,8 +82,8 @@ turndownService.addRule("refshortcode", {
     const ref = turndownService.escape(
       node
         .getAttribute("href")
-        .replace(REFSHORTCODESTART, '{{< ref "')
-        .replace(REFSHORTCODEEND, '" >}}')
+        .replace(REFSHORTCODESTART, '{{% ref "')
+        .replace(REFSHORTCODEEND, '" %}}')
```
aliceriot: just curious, why was this change necessary?

src/markdown_generators.js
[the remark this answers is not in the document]
aliceriot: ahh yeah on HTMLElement you have to do something like [...el].forEach to be able to use it.. tbh I don't know why .forEach isn't implemented there...kind of annoying.

src/markdown_generators.js
gumaerc: It wasn't actually necessary per se, but:

> In Hugo 0.55 we changed how the % delimiter works. Shortcodes using the % as the outer-most delimiter will now be fully rendered when sent to the content renderer

I figured we should standardize on using shortcodes this way, even if it doesn't really matter for the `ref` shortcode.  This means that the output of the shortcode will be rendered as markdown fully before the final markdown processing step.

Issue #14, opened 2020-01-02
Home page menu config at wrong level


gumaerc:
### Steps to Reproduce

Convert some OCW JSON to markdown, then generate a hugo site with hugo-course-publisher.

### Expected Behavior

The Course Home page gets its own menu entry at the top of the list.

### Actual Behavior

The Course Home menu entry does not appear.

PR #15, opened 2020-01-02
Course home menu entry was at the wrong config level

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/14

#### What's this PR do?
Front matter generation for the course home page had the menu item entry inside of the course_info node, when it should have been at the root.

Convert a course, then generate a static site using hugo-course-generator.  The "Course Home" menu item should appear.

### DIFF
--- a/src/lib/markdown_generators.js
+++ b/src/lib/markdown_generators.js
@@ -99,11 +99,11 @@ const generateCourseHomeFrontMatter = courseData => {
       topics:        courseData["course_collections"].map(makeTopic),
       course_number: courseNumber,
       term:          `${courseData["from_semester"]} ${courseData["from_year"]}`,
-      level:         courseData["course_level"],
-      menu:          {
-        main: {
-          weight: -10
-        }
+      level:         courseData["course_level"]
+    },
+    menu: {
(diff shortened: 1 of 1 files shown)


noisecapella:
FYI I added an issue to test for this: https://github.com/mitodl/ocw-to-hugo/issues/17

gumaerc:
> FYI I added an issue to test for this: #17

Sounds good, but I don't think a check like that would be triggered by this.  Hugo doesn't enforce a specific scheme when it comes to the front matter.  If the menu item is there in the front matter, it adds the item to the menu, but if it isn't it just doesn't add it.  The executable will still exit with code 0.

noisecapella:
Ah, I see what you mean. I was getting an error after conversion of a course but it was a missing shortcode error. I think we still need some kind of test for this so we can catch this in the future, though it may be best to figure this out holistically with hugo-course-publisher before we do so.

gumaerc:
Which missing shortcode, and what branch of hugo-course-publisher were you using?

Issue #11, opened 2019-12-30
Links are not processed properly


gumaerc:
### Steps to Reproduce

Convert any course and try to click links inside content.  Certain types of links (i.e. "resolveuid" links) are not being converted.  The code that does this in the ocw-data-parser project does not convert them before putting them into master.json, they are instead converted in the static HTML generator in that project.  This makes sense because another program parsing this data (i.e. Hugo) will expect a different format for links to section pages or static content.

### Expected Behavior

Click a link to another section in the course and it goes there.

### Actual Behavior

Link goes nowhere

PR #16, opened 2020-01-08
Fix links

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/11

#### What's this PR do?
Fixes links in master JSON data before creating markdown

Generate markdown for any number of courses.  All links should point to where they are supposed to go.  There are a few exceptions where a link may exist in the course section HTML that doesn't exist in the course_pages or course_files.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -70,9 +70,10 @@ const getYoutubeEmbedHtml = media => {
     return embeddedMedia["id"] === "Video-YouTube-Stream"
   })
   return youTubeMedia
-    .map(embeddedMedia => {
-      return `<div class="text-center"><iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/${embeddedMedia["media_info"]}" frameborder="0" allow="encrypted-media; picture-in-picture"></iframe></div>`
-    })
+    .map(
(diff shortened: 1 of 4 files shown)


src/helpers.js
```diff
@@ -65,10 +65,22 @@ const makeTopic = feature => {
   return topic
 }
 
+const getYoutubeEmbedHtml = media => {
+  const youTubeMedia = media["embedded_media"].filter(embeddedMedia => {
+    return embeddedMedia["id"] === "Video-YouTube-Stream"
+  })
+  return youTubeMedia
+    .map(embeddedMedia => {
+      return `<div class="text-center"><iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/${embeddedMedia["media_info"]}" frameborder="0" allow="encrypted-media; picture-in-picture"></iframe></div>`
```
aliceriot: this is like a very optional thing, but if you have a one-line function body you can drop the curly braces and the return statement, so you could do `.map(embeddedMedia => <div ... />)` if you wanted

Issue #18, opened 2020-01-08
Add unit tests


noisecapella:
As a developer, I'd like to have unit tests that cover the functionality so I can refactor with more confidence. The ocw-data-parser code is very messy and hacky and this makes it extra important to have these tests so that we can edit code and keep compatibility.

#### Acceptance Criteria:

- [ ] At least one unit test for each function. We can use sinon to stub out the fs functions.

PR #20, opened 2020-01-27
Add unit tests

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/18

#### What's this PR do?
Adds unit tests to assert output of each function

Clone the repo, run `yarn install`, then `npm test`


src/file_operations_test.js
```diff
@@ -0,0 +1,189 @@
...
+  afterEach(() => {
+    sandbox.restore()
+  })
+
+  it("throws an error when you call it with no source directory", () => {
+    try {
+      fileOperations.scanCourses(null, destinationPath)
+    } catch (err) {
+      assert.equal(err.message, "Invalid source directory")
+    }
```
aliceriot: you might be able to use the chai `assert.throws` method for these, although what you have here works too

src/file_operations_test.js
```diff
@@ -0,0 +1,189 @@
...
+      assert.equal(err.message, "Invalid destination directory")
+    }
+  })
+
+  it("calls readdirSync once", () => {
+    fileOperations.scanCourses(sourcePath, destinationPath)
+    assert(
+      readdirSync.calledOnce,
+      "Expected readdirSync to be called once on the source path"
+    )
```
aliceriot: it's not crucial, but `sinon` has some assertion methods that will produce nicer error messages when they fail (iirc it's like `sinon.assert.calledOnce(stub)`)

src/file_operations_test.js
```diff
@@ -0,0 +1,189 @@
...
+  })
+
+  afterEach(() => {
+    sandbox.restore()
+  })
+
+  it("calls mkDirSync to create sections folder", () => {
+    mkDirSync.calledWithExactly(
+      path.join(destinationPath, singleCourseId, "sections")
+    )
```
aliceriot: I think you forgot the assert on this one

src/markdown_generators_test.js
```diff
@@ -0,0 +1,350 @@
...
+    )
+    const expectedSections = singleCourseJsonData["course_pages"]
+      .filter(page => page["text"])
+      .map(page => page["short_url"])
+    const sections = singleCourseMarkdownData.map(section => {
+      return section["name"]
+    })
+    assert(
+      sections.includes("_index.md"),
+      "expected _index.md to be in the markdown data"
```
aliceriot: theres an `assert.include` you may be able to use here

src/markdown_generators_test.js
```diff
@@ -0,0 +1,350 @@
...
+    sandbox.restore()
+  })
+
+  it(`sets the title of the page to "Course Home"`, () => {
+    const expectedValue = "Course Home"
+    const foundValue = courseHomeFrontMatter["title"]
+    assert.equal(
+      expectedValue,
+      foundValue,
+      `expected ${expectedValue} to equal ${foundValue}`
```
aliceriot: I think if you use `assert.equal` the message that it prints will be pretty much what you've provided here by default

src/file_operations.js
```diff
@@ -0,0 +1,96 @@
...
+const readdir = util.promisify(fs.readdir)
+let directoriesScanned = 0
+
+const directoryExists = directory => {
+  return (
+    directory &&
+    fs.existsSync(directory) &&
+    fs.lstatSync(directory).isDirectory()
+  )
+}
```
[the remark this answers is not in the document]
gumaerc: Seriously... it should at least be a part of `fs`

package.json
```diff
@@ -47,6 +47,7 @@
     "prettier-eslint": "^9.0.1",
     "prettier-eslint-cli": "^5.0.0",
     "sinon": "^8.0.4",
+    "sinon-chai": "^3.4.0",
```
aliceriot: ooh, I don't think we're using this elsewhere, might have to add that to open soon...

PR #21, opened 2020-02-11
Hierarchical content generation

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

https://trello.com/c/JkknCyTL/11-hierarchical-left-nav

#### What's this PR do?
Adds support to ocw-to-hugo for generating hierarchical content based on the `uid` and `parent_uid` relationship on `course_pages`.

Run courses through the program, load the resulting markdown file structure into hugo-course-publisher and verify that links work and the site is generated as expected at this point in time.  Also, the tests should pass.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -119,21 +119,21 @@ turndownService.addRule("refshortcode", {
 const fixLinks = (page, courseData) => {
   let htmlStr = page["text"]
   if (htmlStr) {
-    const coursePagesWithText = courseData["course_pages"].filter(
-      page => page["text"]
+    const coursePages = courseData["course_pages"].filter(
+      page => page["type"] !== "CourseHomeSection"
     )
-    coursePagesWithText.forEach(coursePage => {
+    coursePages.forEach(coursePage => {
       const placeholder = new RegExp(
         `\\.?\\/?resolveuid\\/${coursePage["uid"]}`,
         "g"
       )
-      const prefix =
-        page["title"] === "Course Home" && coursePage["title"] !== "Course Home"
-          ? "sections/"
-          : ""
       htmlStr = htmlStr.replace(
         placeholder,
-        `REFSHORTCODESTART${prefix}${coursePage["short_url"]}REFSHORTCODEEND`
+        `REFSHORTCODESTART${helpers.pathToChildRecursive(
+          path.join("courses", courseData["short_url"], "sections"),
+          coursePage,
+          courseData
+        )}REFSHORTCODEEND`
       )
     })
     courseData["course_files"].forEach(media => {
@@ -164,7 +164,9 @@ const generateMarkdownFromJson = courseData => {
   courseHomeMarkdown += generateCourseFeatures(courseData)
   courseHomeMarkdown += generateCourseCollections(courseData)
   const rootSections = courseData["course_pages"].filter(
-    page => page["parent_uid"] === courseData["uid"]
+    page =>
+      page["parent_uid"] === courseData["uid"] &&
+      page["type"] !== "CourseHomeSection"
   )
   let menuIndex = 0
   return [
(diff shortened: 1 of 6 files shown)


aliceriot:
generating courses and rendring them in the publisher seems to work!

src/markdown_generators.js
```diff
@@ -162,31 +163,88 @@ const generateMarkdownFromJson = courseData => {
...
+          children,
+          menuIndex,
+          courseData
+        )
       }
     })
   ]
 }
 
+const generateChildMarkdownRecursive = (
```
aliceriot: it would be sort of nice aesthetically if we could map one function across `rootSections` above and also map that across the children - I see some logic difference in this function in terms of how we calculate paths and stuff, but do you think it could be combined with the above into one recursive function to traverse all the pages? if not no worries, was just thinking it would be nice to be able to represent this transformation as a single function.

src/markdown_generators.js
```diff
@@ -118,21 +119,21 @@ turndownService.addRule("refshortcode", {
...
         "g"
       )
-      const prefix =
-        page["title"] === "Course Home" && coursePage["title"] !== "Course Home"
-          ? "sections/"
-          : ""
       htmlStr = htmlStr.replace(
         placeholder,
-        `REFSHORTCODESTART${prefix}${coursePage["short_url"]}REFSHORTCODEEND`
+        `REFSHORTCODESTART${helpers.pathToChildRecursive(
```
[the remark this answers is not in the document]
gumaerc: Good catch, I think this was a rebase issue.

src/markdown_generators.js
gumaerc: I think I was able to refactor this nicely so it's just a function that can be passed into map on any array of `course_pages` objects.

src/markdown_generators.js
```diff
@@ -159,34 +160,57 @@ const generateMarkdownFromJson = courseData => {
   /**
     This function takes JSON data parsed from a master.json file and returns markdown data
     */
+  this["courseData"] = courseData
+  this["menuIndex"] = 0
```
aliceriot: just a thought I had here, in JavaScript 'basic' types like numbers, booleans, strings are pass-by-value when they are passed to functions as arguments, but objects, by contrast are pass-by-reference. so if you do something like this:

```js
let foo = 0
(x => { x += 1 })(foo)
```

then `foo` will still be `0`, because when the function is called it is not passed a reference to the `foo` variable, but instead a new variable `x` is instantiated with a copy of `foo`'s value, so the mutation of `x` doesn't affect the value of `foo`.

but since objects are pass-by-reference, if you do this:

```js
let fooContainer = { foo: 0 }
(x => { x.foo += 1 })(fooContainer)
``` 

then `fooContainer.foo` will be 1, b/c the function receives a reference to the original `fooContainer` variable and not just a copy of it's value, so modifying `fooContainer.foo` in the function body 'escapes' the function's context and modifies the original object.

Anyhow, I think if you wanted to you could use this to refactor away from using `this` to hold the `menuIndex` variable, if it were passed down as `{ menuIndex: 0 }` for instance I think you could modify it anywhere in the recursive function call tree and have it be visible globally elsewhere. Feel free to ignore if you don't want ot bother, just something I was thinking about as I read :smile:

gumaerc: I knew about by-ref vs by-val arguments, but didn't didn't know some of these specifics about Javascript.  I tried setting the functions up to take courseData and menuIndex as arguments, but I think it made the code a little messier actually.  Since the default arguments coming through map are the iterated object, its index and the source array... you have to use this:

```
rootSections.map(page => {
      generateMarkdownRecursive({
        page:       page,
        courseData: courseData,
        menuIndex:  0
      })
    })

...

children.map(child => {
      generateMarkdownRecursive({
        page: child,
        courseData: data["courseData"],
        menuIndex: data["menuIndex"]
      })
    })
```

versus just setting the initial data on `this` and not even having to worry about arguments:
```
this["courseData"] = courseData
this["menuIndex"] = 0

rootSections.map(generateMarkdownRecursive, this)

...

children.map(generateMarkdownRecursive, this)
```

PR #25, opened 2020-02-25
Course collections speciality

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://trello.com/c/RQRqs9FQ/35-topic-subtopic-specialty

#### What's this PR do?
Removes the `makeTopic` function and refactors `getCourseCollectionsText` to serve as a singular function for getting a text representation of an object from the `course_collections` array in `master.json`.  "Specialty was also changed to "speciality" to match what's in the `master.json` file and the speciality is now being rendered as the last part of the string that is returned.

All tests should pass, and you should be able to run the program against the sample courses and see topics being rendered with their speciality.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -42,18 +42,28 @@ const getCourseSectionFromFeatureUrl = courseFeature => {
   return urlParts[urlParts.length - 1].split("#")[0]
 }
 
-const getCourseCollectionText = (courseCollection, separator) => {
+const getCourseCollectionObject = courseCollection => {
   const feature = courseCollection["ocw_feature"]
   const subfeature = courseCollection["ocw_subfeature"]
   const speciality = courseCollection["ocw_speciality"]
-  let collection = feature
+  const collection = {}
+  if (feature) {
+    collection["feature"] = feature
+  }
   if (subfeature) {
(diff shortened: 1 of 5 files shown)


src/helpers.js
```diff
@@ -42,29 +42,28 @@ const getCourseSectionFromFeatureUrl = courseFeature => {
...
+    collection["topic"] = feature
+  }
   if (subfeature) {
-    collection = `${collection} > ${subfeature}`
+    collection["subtopic"] = subfeature
   }
-  if (specialty) {
-    collection = `${collection} > ${specialty}`
+  if (speciality) {
+    collection["speciality"] = speciality
```
aliceriot: are we thinking we want to stick with the `speciality` spelling going forward? I guess we probably don't want to mix them...

aliceriot: question for @pdpinch too I think

gumaerc: I don't have any strong opinions about the spelling itself.  As you said: tomato, tomahto.  However, I vote that we should keep it `speciality` to be consistent with the way it is in master.json so that if someone in the future is trying to look back at the source data they will be less confused maybe.

aliceriot: sounds good to me

PR #35, opened 2020-03-18
PDF page generation

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://trello.com/c/jK4E1jk9/34-desktop-mobile-pdf-viewer

#### What's this PR do?
This PR adds support for iterating the `course_files` array and rendering PDF files that it finds as markdown pages.  It changes the behavior a bit in that pages that have PDF's attached to them will become sections (folders with an `_index.md` file) and the PDF will be deposited as children of that section, but not added to the nav.

Test should pass, and you should be able to generate markdown from the ocw-next sample courses with hugo-course-publisher.  Note that since this PR contains a new shortcode, you'll need to use the corresponding branch, named `cg/pdf-viewer`.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -330,7 +330,7 @@ const generatePdfMarkdown = (file, courseData) => {
   const pdfFrontMatter = {
     title:         file["title"],
     description:   file["description"],
-    type:          "file",
+    type:          "courses",
     layout:        "pdf",
     file_type:     file["file_type"],
     file_location: file["file_location"],
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -210,7 +210,7 @@ const generateMarkdownRecursive = page => {
   return {
     name:
       isParent || hasFiles
-        ? path.join(pathToChild, "_index.md")
+        ? path.join(pathToChild, "index.md")
         : `${pathToChild}.md`,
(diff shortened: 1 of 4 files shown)


src/markdown_generators.js
```diff
@@ -202,9 +239,18 @@ const generateMarkdownRecursive = page => {
...
+        ? path.join(pathToChild, "_index.md")
+        : `${pathToChild}.md`,
     data:     courseSectionMarkdown,
-    children: children.map(generateMarkdownRecursive, this)
+    children: children.map(generateMarkdownRecursive, this),
+    files:    pdfFiles.map(file => {
+      return {
+        name: `${path.join(pathToChild, file["id"].replace(".pdf", ""))}.md`,
+        data: generatePdfMarkdown(file, courseData)
+      }
```
aliceriot: when you're returning an object literal like this you can just wrap it in `( )` if you want:

```js
file => ({
  name: ...,
  data: ...
})
```

save two lines :laughing:

gumaerc: I think somewhere in the back of my head I knew this.  I'm pretty sure that I started writing the inside of the map by declaring some variables and outputting them with console.log to make sure I was getting the right values.  I consolidated it to one return statement and forgot to take it a step further ;-)

PR #36, opened 2020-03-23
add thumbnail image

abeglova:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://trello.com/c/qiCvN0U8/90-add-thumbnail-to-mobile-course

#### What's this PR do?
1) Adds `course_thumbnail_image_url ` to the hugo markdown

Download dropbox course link with parsed ocw data which includes  `thumbnail_image_src`  in the master json.

Run the ocw to hugo scrip with the new course data. Verify that the script runs correctly and the hugo markdown includes `course_thumbnail_image_url`


src/markdown_generators.js
```diff
@@ -218,6 +218,7 @@ const generateCourseHomeFrontMatter = courseData => {
     course_id:                   courseData["short_url"],
     course_title:                courseData["title"],
     course_image_url:            courseData["image_src"],
+    course_thumbnail_image_url:  courseData["thumbnail_image_src"],
```
abeglova: The script breaks if the `_master.json` does not include `thumbnail_image_src`. This is consistent with the behavior for other desired markdown fields. Does this seem like a problem?

aliceriot: I think maybe that's a quesiton for @gumaerc - any idea if any courses lack a thumbnail? I thought that in ocw-to-hugo we always put a value for these, but that sometimes it's a placeholder if the original data is not there.

abeglova: Once the ocw-data-parser pr is merged, `thumbnail_image_src` will always be populated in the master json, but it could be blank if there is no thumbnail in the course data

aliceriot: in that case I think it's probably ok for us to do this here, but we might want to make ocw-data-parser put in a placeholder if it's not there so we can assume it is here and in hugo-course-publisher?

gumaerc: I was looking around for the logic surrounding this, and I think it has just been removed at this point.  Basically I think it was just a css rule that defaulted to `images\course_image.jpg` as the background image for a div, and if an image came through in the markdown that would be set directly on the template.  The logic has changed since then, so it's simply an `img` tag with the `src` set to `"{{ relURL .CurrentSection.Params.course_image_url }}"`.  So essentially, there is no "default" course image now and we're assuming that it will always come through.  I agree that this is fine to merge for now, but maybe we should decide at what level a placeholder gets inserted at later.

PR #37, opened 2020-03-30
Video pages

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

https://trello.com/c/Kg66zdss/100-create-markdown-files-for-video-pages

#### What's this PR do?
This PR makes ocw-to-hugo generate a markdown file for media listed in the `course_embedded_media` array from master.json.1

Tests should pass.  You should be able to run the program against the sample courses, now also getting back a page for every piece of course media.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -3,8 +3,6 @@
 const path = require("path")
 const yaml = require("js-yaml")
 const markdown = require("markdown-builder")
-const JSSoup = require("jssoup").default
-const titleCase = require("title-case")
 const TurndownService = require("turndown")
 const turndownPluginGfm = require("turndown-plugin-gfm")
 const helpers = require("./helpers")
@@ -223,6 +221,8 @@ const generateMarkdownRecursive = page => {
   )
     .map(key => {
       const embeddedMedia = courseData["course_embedded_media"][key]
+      embeddedMedia["type"] = "courses"
+      embeddedMedia["layout"] = "video"
       return embeddedMedia["parent_uid"] === page["uid"] ? embeddedMedia : null
     })
     .filter(embeddedMedia => embeddedMedia)
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -3,6 +3,7 @@
 const path = require("path")
 const yaml = require("js-yaml")
 const markdown = require("markdown-builder")
+const JSSoup = require("jssoup").default
 const titleCase = require("title-case")
 const TurndownService = require("turndown")
 const turndownPluginGfm = require("turndown-plugin-gfm")
@@ -222,9 +223,6 @@ const generateMarkdownRecursive = page => {
   )
     .map(key => {
       const embeddedMedia = courseData["course_embedded_media"][key]
-      // this currently causes a memory overflow
-      // embeddedMedia["about_this_resource_text"] = turndownService.turndown(embeddedMedia["about_this_resource_text"])
-      // embeddedMedia["transcript"] = turndownService.turndown(embeddedMedia["transcript"])
       return embeddedMedia["parent_uid"] === page["uid"] ? embeddedMedia : null
     })
     .filter(embeddedMedia => embeddedMedia)
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -217,6 +217,13 @@ const generateMarkdownRecursive = page => {
       file["file_type"] === "application/pdf" &&
       file["parent_uid"] === page["uid"]
   )
+  const coursePageEmbeddedMedia = Object.keys(courseData["course_embedded_media"]).map(key => {
+    const embeddedMedia = courseData["course_embedded_media"][key]
+    // this currently causes a memory overflow
+    // embeddedMedia["about_this_resource_text"] = turndownService.turndown(embeddedMedia["about_this_resource_text"])
+    // embeddedMedia["transcript"] = turndownService.turndown(embeddedMedia["transcript"])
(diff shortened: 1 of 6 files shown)


src/markdown_generators_test.js
```diff
@@ -64,6 +65,11 @@ describe("generateMarkdownFromJson", () => {
             return markdownData["name"]
           }
         )
+        const mediaMarkdownFileNames = sectionMarkdownData["media"].map(
+          markdownData => {
+            return markdownData["name"]
+          }
```
aliceriot:
```suggestion
          markdownData => markdownData["name"]
```

src/markdown_generators_test.js
```diff
@@ -72,6 +78,17 @@ describe("generateMarkdownFromJson", () => {
...
+          singleCourseJsonData["course_embedded_media"]
+        )
+          .map(key => {
+            const embeddedMedia =
+              singleCourseJsonData["course_embedded_media"][key]
+            return embeddedMedia["parent_uid"] === sectionUid
+              ? embeddedMedia
+              : null
+          })
+          .filter(embeddedMedia => embeddedMedia)
```
aliceriot: I think here you could also use `Object.values` and then just put the `embeddedMedia["parent_uid"] === sectionUid` check into your `.filter`

src/markdown_generators.js
```diff
@@ -217,11 +216,22 @@ const generateMarkdownRecursive = page => {
       file["file_type"] === "application/pdf" &&
       file["parent_uid"] === page["uid"]
   )
+  const coursePageEmbeddedMedia = Object.keys(
+    courseData["course_embedded_media"]
+  )
```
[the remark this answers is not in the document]
aliceriot: you could also put the `parent_uid` check into the `.filter` call

aliceriot:
there were a few things I'm wondering about as I start to work on building out a basic video UI page that I figured I might as well ask / raise here.

- on the markdown files with type `courses` and layout `video` there is an array of video-related things in the `embedded_media` param. Is it the case that the youtube entry is always the first one? I have looked through a few and it looks like it mostly is, but I'm not sure.
- is seems like on a lot in some courses there is a folder where these video files are ending up and others it is different. I'm just wondering how to go about surfacing them to the user. One idea is that in our `course/section.html` we could add something that would show a list of all video lectures (and PDFs) in the folder 'under' it. I know there are designs for a video gallery to implement later - just wondering how we should go about making these things even visible in the UI right now, since the majority of them seem to be only accessible directly via URL.

pdpinch:
@aliceriot 

1. I’m pretty sure the order of "video-related things" is significant in Plone, so I think you’ll find that YouTube is always first. But it's still data entered by the CMS user without validation. I suppose @gumaerc or @abeglova could do a little analysis to confirm, or we could add some logging for exceptions. 

2. I think I need some example courses of the distinction your mentioning. On ocw.mit.edu there's a couple of different ways video can be authored, but we're trying to start with the canonical "video gallery" in Linear Algebra: https://ocw.mit.edu/courses/mathematics/18-06-linear-algebra-spring-2010/video-lectures/

Some courses don't have a "video gallery" but we may want to create one anyway.

aliceriot:
yeah, for the second thing for instance these two courses have a number of videos:

http://localhost:3000/courses/18-06-linear-algebra-spring-2010/
http://localhost:3000/courses/ec-711-d-lab-energy-spring-2011/

the first one though has a folder in the ocw-to-hugo output called `Video Lectures` where all the video files (i.e. the markdown files representing the video objects) are located, while in the other one the videos are sort of all scattered about.

aliceriot:
in the case of the linear algebra one, it would be great if the `_index.md` file could have a `layout: video` or something along those lines in its frontmatter to signal that we want to render a video gallery there - @gumaerc any ideas about these things? We don't have to by any means handled them in this PR, just a few thoughts I was having as I started working on the video player.

gumaerc:
So, I've addressed the above code problems @aliceriot mentioned.  Looking over these comments, it seems there are two issues:

1. The ordering of embedded media is not well defined
2. We need to set up the data so that pages with videos can use a list template

When Hugo is determining which template to use, the `type` key determines what folder under `layouts` it pulls its layouts from.  Putting `courses` here makes sure that Hugo will use the `baseof.html` from that folder.  Setting `layout` to `video` will make it look specifically for a `video.html` file under the `courses` folder.  I think what we maybe want to do here, for things like PDF's and videos, is create `video` and `pdf` folders under `layouts` and somehow inherit their `baseof.html` template off of the `courses` `baseof.html`.  This way, we could put a `single.html` and a `list.html` under those folders.  This way, for pages that are true video galleries, we could set the type to `video` on `_index.md` and write the `list.html` template under the `layouts/video` folder that would automatically pull all of the markdown files representing individual videos in that section and iterate them.  Then we would have a dedicated layout for rendering a single video page under `layouts/video/single.html`.  Thoughts?

gumaerc:
On the topic of ordering, as I've mentioned in the past I'm fairly certain that the only representation of order in the source JSON (`0.json`, etc.) is literally the number on the file.  There is no "index" property on these files, just the number on the JSON file and they are pulled in in that order.  If we look at this order and it seems to be right & consistent, I could put the number from the filename into each object in `master.json` to indicate order.  That might be the best we can do unless @pdpinch has more insight about where we can obtain some kind of ordering index?

aliceriot:
For the ordering I re-wrote my video page code to use hugo's `where` function to find the appropriate thing to use, so we don't need to worry about the order anymore.

aliceriot:
for the list template yeah, I think that generally makes sense, I was thinking of something generally along those lines

Issue #41, opened 2020-04-09
Pull master json files directly from S3


gumaerc:
As a developer, I'd like to be able to source my master JSON files directly from S3 rather than having to download them locally and put them in a folder.

#### Acceptance Criteria:

- [ ] ocw-to-hugo command accepts a JSON file of a specific format that defines the courses to be downloaded
- [ ] master json files are downloaded from S3 information defined in a `.env` file
- [ ] existing local folder source functionality remains unbroken

PR #42, opened 2020-05-07
AWS Download Functionality

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/41

#### What's this PR do?
The optional `-c` argument has been added, which takes the location of a JSON file containing a structure detailed in the README.  The courses listed in this JSON file are iterated and downloaded to the source folder passed in at `-s`.

Tests should pass, and you should be able to follow the instructions in the README to successfully connect to an `open-learning-course-data` bucket and download a list of courses which are subsequently converted to markdown.


src/file_operations.js
```diff
@@ -21,6 +26,83 @@ const directoryExists = directory => {
   )
 }
 
+const downloadCourses = (coursesJson, coursesDir) => {
```
aliceriot, line 29: I think maybe these two S3 related functions should go in a separate file - perhaps `src/course_download.js` ? it just seems like because they accomplish a fairly specific and separate thing from the rest of the functions in this file it would be better for them to live separately

src/aws_sync.js
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env node
```
aliceriot, line 1: this is typically only necessary in a script we're going to call from the command line - I think we only need it in our `index.js` file

aliceriot:
it's really nice and fast on the repeat download now!

Issue #22, opened 2020-02-19
Relative links to course sections


gumaerc:
### Steps to Reproduce

Generate markdown using the standard sample courses

### Expected Behavior

All links to other course sections function as expected and bring you to that section.

### Actual Behavior

Some links, for example, the "proofs" links in 6-042j-mathematics-for-computer-science-spring-2015 lead to a 404.

These links are relative in the source json and not `resolveuid` links.  We will need a way to recognize and deal with these.

aliceriot:
found a big table here: http://localhost:3000/courses/16-01-unified-engineering-i-ii-iii-iv-fall-2005-spring-2006/sections/fluid-mechanics/ (in hugo-course-publisher) which has a bunch of hardcoded links to PDFs which have the correct path for OCW but not for hugo

PR #43, opened 2020-05-18
Relative links static content

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/22

#### What's this PR do?
The new code in this PR catches relative links in page content and attempts to parse them into their different sections and figure out what to do with them.  Relative links to PDF's are transformed into a getpage shortcode for that PDF's page and other content is rewritten to its `file_location` property on the corresponding item in `course_files`.

Run the conversion.  There should be no errors that crash the program and no pages should silently fail and output HTML into the markdown.


src/helpers.js
```diff
@@ -135,7 +141,153 @@ const pathToChildRecursive = (basePath, child, courseData) => {
...
+          } else {
+            for (let i = parts.length - layers; i < parts.length; i++) {
+              const section = parts[i]
+              if (i + 1 === parts.length) {
+                page = section
+              } else {
+                sections.push(section)
+              }
+            }
+          }
```
aliceriot, line 233: I was thinking, with this logic to break down the URL and fiddle about with the course ID, it may be helpful to any future person reading this code to include a few comments explaining what a typical URL looks like, why we're saying that `layers = parts.length - 4`, etc, because without being familiar with the URLs in question it's pretty opaque what's going on here

src/helpers.js
```diff
@@ -135,7 +141,153 @@ const pathToChildRecursive = (basePath, child, courseData) => {
...
+      courseData
+    )}${getHugoPathSuffix(page, courseData)}`
+    Array.from(htmlStr.matchAll(/\.?\/?resolveuid\/.{0,32}/g)).forEach(
+      match => {
+        const url = match[0]
+        const urlParts = url.split("/")
+        const uid = urlParts[urlParts.length - 1]
+        const linkedPage = courseData["course_pages"].find(
+          coursePage => coursePage["uid"] === uid
+        )
```
aliceriot, line 170: it could be good to add a few comments describing what's going on here too

PR #47, opened 2020-06-18
Some refactoring / renaming and some new tests

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/48

#### What's this PR do?
This PR does a couple things:

- Renames the `-source` and `-destination` arguments to `-input` and `-output`
- Makes the `-courses` argument not automatically trigger downloading of courses, and filters the courses to be converted to markdown based on it as well
- Adds the `-download` boolean argument that tells `ocw-to-hugo` to download courses passed with with `-courses` from AWS
- Adds an `example_courses.json` to the project that can be used as a template
- Adds some tests for the new functionality

Given the tests pass, and assuming you have the following items under the `private` folder:
- `courses.json` - A JSON list of courses for `ocw-to-hugo` to work with, based on `example_courses.json`
- `courses` - A folder containing folders for each course to be processed with the content inside
- `markdown` - A folder to deposit converted courses into

Try the following from the `ocw-to-hugo` source folder.  Each step that generates markdown should be verified by loading the markdown into `hugo-course-publisher` and making sure the site spins up without error.
- [ ] Run `node . -c private/courses.json -d -i private/courses -o private/markdown`, which should download courses from AWS and then convert them to markdown.
- [ ] Run `node . -c private/courses.json -i private/courses -o private/markdown`, which should simply convert the courses in `courses.json` to markdown.
- [ ] Remove a random course from `private/courses` and run the same command as the last step.  You should get a `no such file or directory` error for the missing course, but the process should complete for the other courses.
- [ ] Leaving the `private/courses` folder as it is, run `node . -i private/courses -o private/markdown`.  This should process 54 courses, as there are 55 currently in the example `courses.json` and we removed one.
- [ ] Either put the deleted course back, or run the first command again to re-download it.  Remove the same course from your `courses.json` and run `node . -c private/courses.json -i private/courses -o private/markdown` again.  It should again only process 54 courses, even though the source for the course we removed in still in `private/courses`.


aliceriot:
sorry!

src/bin/index.js
```diff
@@ -5,33 +5,49 @@ const yargs = require("yargs")
...
   })
-  .option("d", {
-    alias:        "destination",
-    describe:     "Destination directory for Hugo markdown",
+  .option("o", {
+    alias:        "output",
+    describe:     "Output directory for Hugo markdown",
     type:         "string",
     demandOption: true
   }).argv
```
aliceriot, line 38: it's too bad `yargs` doesn't support things like a `dir` and `file` type, which could automatically check if they're present and readable / writable

src/file_operations.js
```diff
@@ -5,74 +5,113 @@ const util = require("util")
...
+      coursesJson["courses"].forEach(course =>
+        scanCourse(inputPath, outputPath, course).then(() => {
+          progressBar.increment()
+        })
+      )
+    } else {
+      console.log(NO_COURSES_FOUND_MESSAGE)
     }
-  })
+  } else {
```
aliceriot, line 49: the contents of this `else` look pretty close to the body of the `if (jsonPath)` - maybe we could use the conditional just to get the course list, and then use the same logic to go through it?

something like

```js
const courseList = jsonPath
  ? JSON.parse(fs.readFileSync(jsonPath))["courses"]
  : fs.readDirSync(inputPath)

const numCourses = jsonPath
  ? courseList.length
  : courseList.filter(file =>
      directoryExists(path.join(inputPath, file))
    ).length
```

then I think you could from that point on treat them the same, if I'm reading things correctly

gumaerc: A very good point!  What you posted worked great.

src/bin/index.js
gumaerc: Yea, unfortunately it only supports array, boolean or string

PR #50, opened 2020-07-13
no left nav for grandchildren

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://trello.com/c/IHLhRE3H/159-dont-include-third-level-pages-in-the-left-nav-unless-listinleftnav-is-set-to-true

#### What's this PR do?
This PR adds the `isGrandChild` and `listInLeftNav` argument to `generateCourseSectionFrontMatter`.  Any pages 3rd tier or higher will now not be added to the left nav unless `lisntInLeftNav` is true.

Find a course in that has 3rd tier nav items, for example:

https://ocw-next.netlify.app/courses/res-15-003-shaping-the-future-of-work-15-662x-spring-2016/sections/introduction-challenges-and-opportunities/

Download and parse said course using `ocw-data-parser` on [this)(https://github.com/mitodl/ocw-data-parser/tree/cg/list-in-left-nav) branch.  Use the master json produced from this with this branch of `ocw-to-hugo`, then verify the output has the appropriate nav items removed.

Other relevant PR's:
https://github.com/mitodl/ocw-data-parser/pull/41
https://github.com/mitodl/hugo-course-publisher/pull/149

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -351,7 +351,7 @@ const generateCourseSectionFrontMatter = (
     course_id: courseId
   }
 
-  if (!isGrandChild || (isGrandChild && listInLeftNav)) {
+  if (!isGrandChild || listInLeftNav) {
     courseSectionFrontMatter["menu"] = {
       [courseId]: {
         identifier: pageId,
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -190,9 +190,11 @@ const generateMarkdownRecursive = page => {
   const hasMedia = coursePageEmbeddedMedia.length > 0
   const hasParent = parents.length > 0
   const parent = hasParent ? parents[0] : null
-  const isGrandChild = hasParent ? courseData["course_pages"].filter(
-    coursePage => coursePage["uid"] === parent["parent_uid"]
-  ).length > 0 : false
+  const isGrandChild = hasParent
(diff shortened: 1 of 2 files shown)


src/markdown_generators.js
```diff
@@ -329,27 +336,33 @@ const generateCourseSectionFrontMatter = (
...
     Generate the front matter metadata for a course section given a title and menu index
     */
   const courseSectionFrontMatter = {
     title:     title,
-    course_id: courseId,
-    menu:      {
+    course_id: courseId
+  }
+
+  if (!isGrandChild || (isGrandChild && listInLeftNav)) {
```
aliceriot, line 354: I think this is equivalent to `!isGrandChild || listInLeftNav`, because whenever `isGrandChild` is false then we're doing

```js
!false || ...
```

and we don't even evaluate the expression after the `||`, but conversely when `isGrandChild` is `true` then we're basically doing

```js
!true || (true && listInLeftNav)
```

and `true && x` is always going to be `x`, so the expression `isGrandChild && listInLeftNav` will I think in all cases where it is evaluated be equal to the value of `listInLeftNav` anyhow.

Is that right?

gumaerc: Yes, you're right.  Thank you for the detailed breakdown!

PR #59, opened 2020-08-13
fix escaping of square brackets in link text

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/hugo-course-publisher/issues/166

#### What's this PR do?
The text of some links in OCW content have pre-escaped square brackets in them.  Since text needs to be escaped before going in the square brackets of a markdown link, feeding these pre-escaped square brackets into that results in `\\\[` which is interpreted as LaTeX by MathJax.  This PR un-escapes square brackets before escaping the entire text content of links.

Run a conversion and look at the markdown for the page in the attached issue or any other page that has link text on OCW with square brackets in it and ensure that the resulting markdown has inner square brackets properly escaped and they aren't `\\\[`.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -117,7 +117,7 @@ turndownService.addRule("codeblockfix", {
 /**
  * Build links with Hugo shortcodes to course sections
  **/
-turndownService.addRule("refshortcode", {
+turndownService.addRule("getpageshortcode", {
   filter: (node, options) => {
     if (node.nodeName === "A" && node.getAttribute("href")) {
       if (node.getAttribute("href").includes(GETPAGESHORTCODESTART)) {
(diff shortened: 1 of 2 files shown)


src/markdown_generators.js
```diff
@@ -109,7 +109,10 @@ turndownService.addRule("refshortcode", {
     const children = Array.prototype.slice.call(node.childNodes)
     if (!children.filter(child => child.nodeName === "IMG").length > 0) {
       // if this link doesn't contain an image, escape the content
-      content = turndownService.escape(content)
+      // except first make sure there are no pre-escaped square brackets
+      content = turndownService.escape(
+        content.replace(/\\\[/g, "[").replace(/\\\]/g, "]")
+      )
```
aliceriot, line 115: should we add a quick test for this? could use for instance the link from the economics course that is causing a problem (https://github.com/mitodl/hugo-course-publisher/issues/166)

gumaerc: @aliceriot test added

Issue #62, opened 2020-08-26
escaped ` characters are leaking into markdown


gumaerc:
### Steps to Reproduce

Convert any course with pre-escaped backticks (`1-124j-foundations-of-software-engineering-fall-2000` is one of them) and try and run it with `hugo-course-publisher`.

### Expected Behavior

The site builds normally and you are able to browse it

### Actual Behavior

You get an error about an unrecognized escape character and the Hugo build fails

PR #63, opened 2020-08-26
handle pre-escaped backticks

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/62

#### What's this PR do?
Adds a turndown rule to handle inline code blocks and a helper function to handle pre-escaped backticks

Convert courses with pre-escaped backticks (i.e. `1-124j-foundations-of-software-engineering-fall-2000`) and make sure `hugo-course-publisher` builds properly with it in the content directory.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -118,7 +118,7 @@ turndownService.addRule("codeblockfix", {
 /**
  * turn kbd, tt and samp elements into inline code blocks
  */
-turndownService.addRule("kbdfix", {
+turndownService.addRule("inlinecodeblockfix", {
   filter: node =>
     node.nodeName === "KBD" ||
     node.nodeName === "TT" ||
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -135,14 +135,13 @@ turndownService.addRule("inlinecodeblockfix", {
       content.match(backTickSingleQuoteWrap).forEach(match => {
         content = content.replace(`\\\`${match}'`, match)
       })
-      return `\`${content}\``
     } catch (err) {
       loggers.fileLogger.log({
         level:   "error",
         message: err
       })
     }
-    return content
+    return `\`${content}\``
   }
 })
 
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -116,6 +116,35 @@ turndownService.addRule("codeblockfix", {
 })
 
 /**
+ * turn kbd, tt and samp elements into inline code blocks
+ */
+turndownService.addRule("kbdfix", {
+  filter: node =>
+    node.nodeName === "KBD" ||
+    node.nodeName === "TT" ||
+    node.nodeName === "SAMP",
+  replacement: (content, node, options) => {
+    /**
+     * fix weird formatting issue with some elements' content being wrapped
+     * between a backtick and a single quote
+     */
+    try {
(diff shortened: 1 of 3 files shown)


aliceriot:
I may have something messed up locally, but I pulled down the software course and ran it through `ocw-to-hugo` with this branch and I've gotten something rather strange out the other end. When I copy that course into `hugo-course-publisher` it seems that there's a few weird things, using that Makefile docs page as an example.

One thing is a whole bunch of anchors at the beginning:

Screenshot from 2020-09-01 16-46-25
Screenshot from 2020-09-01 16-46-06

then it looks also like some things that should be multi-line code blocks are being converted to simple paragraphs in ocw-to-hugo:

Screenshot from 2020-09-01 16-47-24

not sure what we want to do about this, or how important this class is. I don't imagine a lot of people these days taking online programming courses are starting with the intricacies of GNU Make...

gumaerc:
Hmm, so the anchor tag thing has to do with `a` tags in the HTML for that file that have a `name` attribute but also actual link text.  The anchor shortcode doesn't apply any text because OCW seems to only use them without text as literal anchors, but another PR could be made to have it pass text.  As for the backtick situation, that's a bit messier.  I think we should definitely either get rid of or update that makefile documentation because as you noted, it's copied and pasted in from somewhere else.  There are pre-escaped backticks in other courses though, and they completely break the hugo build if there are any present in markdown.  I'll have to take another look at this and think about it again.

gumaerc:
So I took another look at this today, and my new branch based off master can build all 3180 courses without these changes to strip out backticks.  I'm not sure why I was having so much trouble earlier, but this entire PR might not be necessary now.

gumaerc:
Yup, closing this as there seemed to be some kind of data corruption issue on my PC that made me think this was an issue but it's not.  That GNU makefile documentation should just be externally linked and not embedded into OCW.

gumaerc:
Re-opening this as it has reared its head again.

aliceriot:
I'm still seeing that issue with the anchors:

Screenshot from 2020-09-15 15-56-28

should we not be concerned about that?

gumaerc:
@aliceriot I forgot to make an issue for that when we talked about this last because this issue was closed, but I'll make one to fix that separately.  The issue there is just that when the anchor shortcode is written, only the `name` attribute is being written to the page.  It needs to be modified to also pass inner HTML content, but that's a separate issue.  Either way, that makefile documentation should be removed and externally linked as we had discussed prior.

gumaerc:
@aliceriot : https://github.com/mitodl/ocw-to-hugo/issues/75

src/markdown_generators.js
```diff
@@ -115,6 +115,36 @@ turndownService.addRule("codeblockfix", {
...
+    node.nodeName === "KBD" ||
+    node.nodeName === "TT" ||
+    node.nodeName === "SAMP",
+  replacement: (content, node, options) => {
+    /**
+     * fix weird formatting issue with some elements' content being wrapped
+     * between a backtick and a single quote
+     */
+    try {
+      // eslint seems to think the escaped backtick in the regex is useless, but it's not
```
aliceriot, line 132: when your linter gets it wrong :upside_down_face:

aliceriot (approved):
other than that thing I think this looks good

PR #66, opened 2020-08-27
mirror hugo-course-publisher contentDir

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

This is part of the work for https://github.com/mitodl/hugo-course-publisher/issues/184
Closes https://github.com/mitodl/hugo-course-publisher/issues/191
Closes https://github.com/mitodl/ocw-to-hugo/issues/67

#### What's this PR do?
This adds code to fetch the `hugo-course-publisher` content boilerplate and inject it into the output directory along with courses.  This is so we can use the direct output of `ocw-to-hugo` as the `contentDir` argument in hugo when building `hugo-course-publisher`.

Run a markdown conversion and then build the site using `hugo-course-publisher` with the following commands:

```sh
npm run build:webpack
hugo -d dist -s site -v --contentDir (ocw-to-hugo output folder here)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,12 @@
     "dotenv": "^8.2.0",
     "js-yaml": "^3.13.1",
     "markdown-builder": "^0.9.0",
+    "simple-git": "^2.20.1",
     "title-case": "^3.0.2",
+    "tmp": "^0.2.1",
     "turndown": "^6.0.0",
     "turndown-plugin-gfm": "^1.0.2",
+    "walk": "^2.3.14",
     "winston": "^3.2.1",
     "yargs": "^15.0.2"
   },
@@ -54,7 +57,6 @@
     "prettier-eslint-cli": "^5.0.0",
     "rimraf": "^3.0.2",
     "sinon": "^8.0.4",
-    "sinon-chai": "^3.4.0",
-    "tmp": "^0.1.0"
+    "sinon-chai": "^3.4.0"
   }
 }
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -29,9 +29,12 @@
     "dotenv": "^8.2.0",
     "js-yaml": "^3.13.1",
     "markdown-builder": "^0.9.0",
+    "simple-git": "^2.20.1",
     "title-case": "^3.0.2",
+    "tmp": "^0.2.1",
     "turndown": "^6.0.0",
     "turndown-plugin-gfm": "^1.0.2",
+    "walk": "^2.3.14",
```
aliceriot, line 37: it looks like the `simple-git` and `walk` packages we're not using anywhere

src/constants.js
```diff
@@ -10,5 +10,19 @@ module.exports = {
...
+      name: "_index.md",
+      content: `---\ntitle: Hugo Course Publisher\n---\n`
+    }, {
+      path: "search",
+      name: "_index.md",
+      content: `---\ntitle: Search\ntype: search\n---\n`
+    }, {
+      path: "courses",
+      name: "_index.md",
+      content: `---\ntitle: Courses\ntype: courseindex\n---\n`
```
aliceriot, line 25: one thing I might suggest is storing the contents as objects here - like `content: { title: "Hugo Course Publisher" }` and then using `yaml.safeDump` to convert them when we write the files - I'm just thinking that will make it a little easier to edit these values in the future if we need to add a property

aliceriot (changes requested):
just one question really

src/constants.js
gumaerc: good idea!

aliceriot (approved):
LGTM :+1:

Issue #68, opened 2020-09-02
Enforcement of AWS access keypair environment variables does not work with EC2 instance profile


markbreedlove:
The application currently requires the `AWS_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY` variables to be set in `.env`, but these are not available if you are running an EC2 instance with an IAM Instance Profile. When running with an instance profile, you don't have an access key pair to maintain. The application should allow for these not to be set.  If the Node version of the `aws` client library works like the Python one, it is able to function automagically on an EC2 instance with an instance profile without these environment variables.

gumaerc:
I didn't know about this, thanks.

PR #72, opened 2020-09-03
AWS config fixes

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/68

#### What's this PR do?
Changes the way aws_sync connects to AWS to support Netlify and theoretically instance configs, although I have no way to test that myself.

Run `ocw-to-hugo` with the `-d` and `-c` flags with the AWS config variables from the readme set in a `.env` file and also with no `.env` file and the values only set in the environment.

### DIFF
--- a/src/aws_sync.js
+++ b/src/aws_sync.js
@@ -3,10 +3,10 @@
 const fs = require("fs")
 const path = require("path")
 const AWS = require("aws-sdk")
-const env = require("dotenv").config().parsed
+require("dotenv").config()
 const cliProgress = require("cli-progress")
 
-const { directoryExists, createOrOverwriteFile } = require("./helpers")
+const { createOrOverwriteFile } = require("./helpers")
 
 const progressBar = new cliProgress.SingleBar(
   { stopOnComplete: true },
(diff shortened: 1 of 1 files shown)


markbreedlove:
If I leave out `AWS_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY` from the `.env` file, I still get an error saying `AWS credentials not set`, on line 46 of the new commit of `aws_sync.js`.

The key variables will never be set on an instance that's using an EC2 instance profile.

```
ocw@ip-10-13-2-88:~/ocw-to-hugo$ ocw-to-hugo -c example_courses.json -d -i ~/testdir-input -o ~/testdir-output
(node:8504) UnhandledPromiseRejectionWarning: Error: AWS credentials not set
    at downloadCourses (/home/ocw/ocw-to-hugo/src/aws_sync.js:46:13)
    at run (/home/ocw/ocw-to-hugo/src/bin/index.js:42:11)
    at Object.<anonymous> (/home/ocw/ocw-to-hugo/src/bin/index.js:53:1)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
(node:8504) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:8504) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

gumaerc:
@markbreedlove Interesting, so `AWS_REGION` would be set then?  I put a comment in there, but the reason this has to be done is that for development it's really nice to be able to just have the environment in a `.env` file, but the AWS SDK doesn't seem to work with this, even after `dotenv` has been initialized and the values from `.env` have been deposited into `process.env`.  I just pushed up a change so all of the variables will be required to be set before it tries to use them, so you shouldn't get an error if they are *all* not set.

markbreedlove:
Thanks, the download works now. `AWS_REGION` and `AWS_BUCKET_NAME` should still be set, but it was just the key variables that had to be made optional.

src/aws_sync.js
```diff
@@ -20,21 +20,34 @@ const downloadCourses = async (coursesJson, coursesDir) => {
...
+    process.env["AWS_REGION"] &&
+    process.env["AWS_ACCESS_KEY"] &&
+    process.env["AWS_SECRET_ACCESS_KEY"]
   ) {
-    throw new Error("Invalid AWS connection info")
+    const config = {}
+    if (process.env["AWS_REGION"]) {
+      config["region"] = process.env["AWS_REGION"]
+    }
+    if (process.env["AWS_ACCESS_KEY"] && process.env["AWS_SECRET_ACCESS_KEY"]) {
```
aliceriot, line 40: it looks like this `if` block (and the one above on line 37) is redundant because of the one on line 31

aliceriot (changes requested):
tested it and everything seems to work just fine, just one question

src/aws_sync.js
gumaerc: Ah yea, I flipped the logic on it to require all of them to be set and forgot about that...

aliceriot (approved):
LGTM :+1:

Issue #73, opened 2020-09-11
ocw-to-hugo throws an error on unpublished courses


gumaerc:
### Steps to Reproduce

Run `ocw-to-hugo` against the data in `open-learning-course-data-production`, the entire bucket.

### Expected Behavior

All courses are processed with no errors

### Actual Behavior

Tons of errors spit out into the console if there are folders with no master.json files in them

PR #74, opened 2020-09-11
don't throw an error to the console if a course is missing

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/73

#### What's this PR do?
Prevents `ocw-to-hugo` from throwing an error to the console when there is a folder in the input directory without a master.json file and only logs it to the file error log instead.

Place a folder in your input folder with no master.json file in it.  When run, `ocw-to-hugo` should not spit an error out to the console and instead just write it to the error log.

### DIFF
--- a/src/file_operations.js
+++ b/src/file_operations.js
@@ -127,7 +127,6 @@ const getMasterJsonFileName = async coursePath => {
     message: courseError
   })
   progressBar.increment()
-  throw new Error(courseError)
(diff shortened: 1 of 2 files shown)


aliceriot (approved):
works great! :+1:

PR #80, opened 2020-09-18
If an error occurs, print the error and exit with a non-zero code

noisecapella:
#### Pre-Flight checklist

- [x] Testing

None

#### What's this PR do?
If an error occurred, the process would hang indefinitely. This catches the error that's thrown, prints it to stderr, and exits with a non-zero exit code.

Run `./src/bin/index.js -i ./test_data/courses -o ./test_data/output -c ./test_data/courses.json --download` and have `AWS_ACCESS_KEY` be empty or set to something that won't work as an access key. You should see the error object with a complaint about the access key, and the process should exit with a non-zero code.

#### Where should the reviewer start?
(Optional)

(Optional)

(Optional)

### DIFF
--- a/src/bin/index.js
+++ b/src/bin/index.js
@@ -69,5 +69,9 @@ const run = async () => {
 
 run().catch(err => {
   console.error("Error:", err)
+  loggers.fileLogger.log({
+    level:   "error",
+    message: err
+  })
   process.exit(1)
 })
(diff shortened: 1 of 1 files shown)


gumaerc (changes requested):
I tested it and it works great, but we also have Winston file logging set up on this project.  Could you also set this up to log to a file like this:

```
loggers.fileLogger.log({
      level:   "error",
      message: err
    })
```

gumaerc (approved):
:+1: LGTM

Issue #75, opened 2020-09-16
anchor shortcodes don't pull in inner HTML content


gumaerc:
### Steps to Reproduce

Visit a page with anchor tags that are being rendered by the anchor shortcode that are supposed to also have inner HTML content, like http://localhost:3000/courses/1-124j-foundations-of-software-engineering-fall-2000/sections/lecture-notes/gnu_makefile_documentation/

### Expected Behavior

The table of contents on the above page renders properly

### Actual Behavior

No text is rendered for the list items

### Screenshot or Screencast

![image](https://user-images.githubusercontent.com/12089658/93373799-d89e0580-f823-11ea-9b68-909cb9555f6f.png)

PR #81, opened 2020-09-21
anchor shortcode content

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/75

#### What's this PR do?
This PR adds `content` as the inner HTML of the generated `anchor` shortcode, which is then pulled in as `.Inner` in the shortcode itself.

Generate some markdown, ensure that `anchor` shortcodes have content in them where they're supposed to.  The turndown rule catches all `a` tags with a `name` property so you should check source HTML that follows this pattern. `1-124j-foundations-of-software-engineering-fall-2000` is a good place to start.

### DIFF
--- a/src/markdown_generators_test.js
+++ b/src/markdown_generators_test.js
@@ -480,7 +480,7 @@ describe("turndown service", () => {
   it("should generate an anchor shortcode for an a tag with a name attribute", () => {
     const inputHTML = `<a name="test">test</a>`
     const markdown = markdownGenerators.turndownService.turndown(inputHTML)
-    assert.equal(markdown, `{{< anchor "test" >}}test{{< anchor />}}`)
+    assert.equal(markdown, `{{< anchor "test" >}}test{{< /anchor >}}`)
   })
 
   it("should turn inline code blocks into text surrounded by backticks", () => {
--- a/src/markdown_generators_test.js
+++ b/src/markdown_generators_test.js
@@ -483,6 +483,15 @@ describe("turndown service", () => {
     assert.equal(markdown, `{{< anchor "test" >}}test{{< /anchor >}}`)
   })
 
+  it("should generate an anchor shortcode for an a tag with a name and href attribute", () => {
+    const inputHTML = `<a name="test" href="https://ocw.mit.edu">test</a>`
+    const markdown = markdownGenerators.turndownService.turndown(inputHTML)
+    assert.equal(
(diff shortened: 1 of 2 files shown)


aliceriot:
I generated the markdown for the 'foundations of software engineering' course and then checked out https://github.com/mitodl/hugo-course-publisher/pull/218/ to render it. I'm now seeing content for all of those anchor things at the beginning of the GNU Makeifle docs:

Screenshot from 2020-09-23 09-47-36

the problem is that most of them aren't clickable - the contents are all `<a>` tags which have only a `name` attribute, where it seems that they should probably have that same value as their `href`. if you look at the current version of the page on OCW prod clicking on any of those section names takes you to the section: https://ocw.mit.edu/courses/civil-and-environmental-engineering/1-124j-foundations-of-software-engineering-fall-2000/lecture-notes/gnu_makefile_documentation/

aliceriot:
I think the problem is that in the original HTML the contents entries have both an `href` attribute and a `name` attribute. so I think in this code we need to grab both of those attributes if they're present on the node.

gumaerc:
> I think the problem is that in the original HTML the contents entries have both an `href` attribute and a `name` attribute. so I think in this code we need to grab both of those attributes if they're present on the node.

Whoops!  `href` is obviously an important part of a link, idk how I missed that.

src/markdown_generators_test.js
```diff
@@ -478,9 +478,9 @@ describe("turndown service", () => {
   })
 
   it("should generate an anchor shortcode for an a tag with a name attribute", () => {
-    const inputHTML = `<a name="test"></a>`
+    const inputHTML = `<a name="test">test</a>`
     const markdown = markdownGenerators.turndownService.turndown(inputHTML)
-    assert.equal(markdown, `{{< anchor "test" >}}`)
+    assert.equal(markdown, `{{< anchor "test" >}}test{{< /anchor >}}`)
```
aliceriot, line 483: kind of silly - but maybe good to just have a quick test that it pulls out an `href` too?

aliceriot (approved):
tried this out with the accompanying hugo-course-publisher pr, works great!

src/markdown_generators_test.js
gumaerc: done

Issue #82, opened 2020-09-21
Remove path exist checks when not needed


noisecapella:
Checking whether a path or file can introduce a [race condition](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) when the status of the directory or file changes between the existence check and when it's being used. Instead, we should just attempt the action anyway and handle the error

PR #83, opened 2020-09-22
Refactor for async/await and error handling

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
In general, error handling is better handled and some `fs` functions are converted to use promises:
 - `fsPromises` is created which wraps `fs.promises` with a function wrapper to create a stack trace for easier tracking of file system errors. All filesystem "sync" functions should be removed from the project in favor of these promise variants. I tried to make these changes as 1-to-1 as possible for easy review.
 - `existsSync` has no promise equivalent because it was deprecated. Instead I added `fileExists` in `helpers.js`. We should get rid of almost all of these existence checks anyway, for security and performance reasons (#82). 
 - Certain file error logging was removed just for redundancy's sake. Errors should generally be thrown and propagated, and then caught and reported in the run function. Other file error logging should still be there in cases where it's not appropriate to throw an error.
 - If any errors were logged in the file logger, a non-zero exit code will be returned after the process completes and written to stderr. The logger was modified to store the logs in memory.
 - `generateCourseSectionMarkdown` was modified to prevent null from being passed to turndown, which caused an error message
(see other notes in comments on this PR)

I'm not sure how this affects performance, but there are some tradeoffs between error handling and performance. We should prioritize error handling wherever possible. There are probably changes we can do to make the project more efficient without removing error handling.

Run `./src/bin/index.js -i ./test_data/courses -o ./test_data/output -c ./test_data/courses.json` on master and on this PR and compare the output directories. There should only be one difference, a "null" which was written out. I think this is a bug which is fixed but let me know if that "null" string should be there.

### DIFF
--- a/src/bin/index.js
+++ b/src/bin/index.js
@@ -65,10 +65,7 @@ const run = async () => {
 run()
   .catch(err => {
     console.error("Error:", err)
-    loggers.fileLogger.log({
-      level:   "error",
-      message: err.message
-    })
+    loggers.fileLogger.error(err)
     process.exit(1)
   })
   .then(() => {
--- a/src/bin/index.js
+++ b/src/bin/index.js
@@ -52,18 +52,13 @@ const run = async () => {
   if (options.courses && options.download) {
     await downloadCourses(options.courses, options.input)
   } else if (options.download && !options.courses) {
-    loggers.fileLogger.log({
-      level:   "error",
-      message: MISSING_JSON_ERROR_MESSAGE
-    })
     throw new Error(MISSING_JSON_ERROR_MESSAGE)
   }
-  writeBoilerplate(options.output).then(() => {
-    scanCourses(options.input, options.output, {
-      courses:      options.courses,
-      strips3:      options.strips3,
-      staticPrefix: options.staticPrefix
-    })
+  await writeBoilerplate(options.output)
+  await scanCourses(options.input, options.output, {
+    courses:      options.courses,
+    strips3:      options.strips3,
+    staticPrefix: options.staticPrefix
   })
 }
 
@@ -71,7 +66,7 @@ run().catch(err => {
   console.error("Error:", err)
   loggers.fileLogger.log({
     level:   "error",
-    message: err
+    message: err.message
   })
   process.exit(1)
 })
--- a/src/bin/index.js
+++ b/src/bin/index.js
@@ -63,10 +63,15 @@ const run = async () => {
 }
 
 run().catch(err => {
-  console.error("Error:", err)
+  console.error("Error:", err, Object.keys(err))
   loggers.fileLogger.log({
     level:   "error",
     message: err.message
   })
   process.exit(1)
+}).then(() => {
+  if (loggers.memoryTransport.logs.length) {
+    console.error("Found errors which were logged: ", loggers.memoryTransport.logs)
+    process.exit(1)
+  }
 })
--- a/src/bin/index.js
+++ b/src/bin/index.js
@@ -46,6 +46,11 @@ const options = yargs
       "When strips3 is set to true, the value passed into this argument will replace the s3 prefix on static assets",
     type:         "string",
     demandOption: false
+  })
+  .option("verbose", {
+    describe:     "Write error logging and other extra output",
+    type:         "boolean",
+    demandOption: false
   }).argv
 
 const run = async () => {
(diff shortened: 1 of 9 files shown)


src/bin/index.js
```diff
@@ -52,26 +52,31 @@ const run = async () => {
...
+    loggers.fileLogger.log({
+      level:   "error",
+      message: err.message
+    })
+    process.exit(1)
+  })
+  .then(() => {
+    if (loggers.memoryTransport.logs.length) {
+      console.error(
+        "Found errors which were logged: ",
```
gumaerc, line 77: I'm not sure it's really necessary to log all these errors to console here.  There is a lot of problematic HTML we've yet to work through and the errors being swallowed up by turndown overflow my terminal buffer with even just the example courses.  With the full course load, this is going to dump a lot of text into the console log.  We should potentially log these to Sentry instead so they can be categorized in a sane manner.  There's no need to clog up the console with these, especially since you can't even scroll back and read all of them.

src/helpers.js
```diff
@@ -17,30 +17,39 @@ const distinct = (value, index, self) => {
...
-const findDepartmentByNumber = departmentNumber => {
-  return departmentsJson.find(department => {
-    return department["depNo"] === departmentNumber.toString()
-  })
+const createOrOverwriteFile = async (file, body) => {
+  const dirName = path.dirname(file)
+  if (!(await directoryExists(dirName))) {
+    await fsPromises.mkdir(dirName, { recursive: true })
+  } else if (await fileExists(file)) {
+    await fsPromises.unlink(file)
```
gumaerc, line 43: When running the conversion subsequent times after the first with the `--download` flag turned on, I get an error from this line:

```sh
Error: Error: Error: ENOENT: no such file or directory, '
    at Object.unlink (/Users/c4103/Github/ocw-to-hugo/src/fsPromises.js:8:13)
    at createOrOverwriteFile (/Users/c4103/Github/ocw-to-hugo/src/helpers.js:43:22)
    at async writeS3Object (/Users/c4103/Github/ocw-to-hugo/src/aws_sync.js:88:5)
    at async Promise.all (index 3)
    at async downloadCourseRecursive (/Users/c4103/Github/ocw-to-hugo/src/aws_sync.js:92:3)
    at async /Users/c4103/Github/ocw-to-hugo/src/aws_sync.js:56:7
    at async Promise.all (index 24)
    at async downloadCourses (/Users/c4103/Github/ocw-to-hugo/src/aws_sync.js:50:10)
    at async run (/Users/c4103/Github/ocw-to-hugo/src/bin/index.js:53:5) {
  message: [Error: ENOENT: no such file or directory, unlink 'private/courses/6-033-computer-system-engineering-spring-{
    errno: -2,
    code: 'ENOENT',
    syscall: 'unlink',
    path: 'private/courses/6-033-computer-system-engineering-spring-2018/0361d9ecba6ae41417b214597ecbe5ac_r2_-2KW76ec.s'
  }
}
```

gumaerc (changes requested):
I was able to run this and generate usable markdown, and generally the new `async` / `await` structure and error handling is nice, but I had a couple issues:

src/bin/index.js
noisecapella: I would like to have this information for debugging purposes. What do you think about leaving it out by default and having this output be enabled with a console argument, `--debug` or similar?

src/helpers.js
noisecapella: Thanks, I'll investigate. I also found a different bug I'm investigating.

src/bin/index.js
gumaerc: That's a great idea, either `--debug` or `--verbose` would be fine.

src/helpers.js
gumaerc: I added the `aws_sync` class so we could maybe have finer control over what gets synced that you can achieve with arguments into `aws s3 sync` but maybe at some point it's worth discussing just using `aws s3 sync` instead to avoid extra issues.  It is convenient to be able to just use the `--download` flag, there are benefits and tradeoffs.

noisecapella:
I fixed a couple other bugs in this PR:
 - `page["short_page_title"]` was sometimes undefined which broke the yaml parser https://github.com/mitodl/ocw-to-hugo/pull/83/commits/ae1ef614b8a88b1df160aab13e9c7369ccbb1814
 - I added a check to fix errors when this regex failed to match https://github.com/mitodl/ocw-to-hugo/pull/83/commits/3d53868d01fede9e8b91143a6b9f4c4f2da78fbc

src/helpers.js
noisecapella: @gumaerc Can you try this again and see if you still see the error?

gumaerc: all fixed :+1:

gumaerc (approved):
:+1: LGTM, although it could use a rebase before merging.  It's probably not super necessary, but most of the markdown differences I noticed were from merged PR's.  The only other difference is that sections without HTML body now don't render "null" in the body which is a net positive.

PR #85, opened 2020-09-22
table header rows fix

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/hugo-course-publisher/issues/219

#### What's this PR do?
This PR changes the turndown rule handling tables and how it renders header columns.  Instead of separating them out and recreating the table, we render the header into the first cell and wrap it in a shortcode which renders the header in a div with the class `position-absolute`.  Then, the other cells in the row are rendered with only a `nbsp;` HTML character entity so that the line height is set correctly.  The column sizes remain consistent, and the header text is allowed to flow into the other columns.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -64,34 +64,20 @@ turndownService.addRule("table", {
     if (content.match(/(?<=\| \*\*)(.*?)(?=\*\* \|\r?\n|\r)/g)) {
       // Get the amount of columns by matching three hyphens and counting
       const totalColumns = content.match(/---/g || []).length
-      // Split headers out on their own so they aren't in one cell
-      return (
-        content
-        /**
-           * First, replace pipe space and double asterisk with a newline
-           * followed by double asterisk
-           */
-
-          .replace(/\| \*\*/g, "\n**")
-          /**
-           * Second, replace double asterisk space pipe followed by a newline
-           * or carriage return with double asterisk double newline.  After the
-           * second newline, re-initialize the table by iterating a pipe followed
-           * by a space and three hyphens for the total amount of columns, finally
-           * closing it out with a single pipe at the end
-           */
-          .replace(
-            /\*\* \|\r?\n|\r/g,
-            `**\n\n${"| ".repeat(totalColumns)}|\n${"| --- ".repeat(
-              totalColumns
-            )}|`
-          )
-          /**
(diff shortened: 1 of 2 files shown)


pdpinch:
Before we go changing how the tables headers are rendered, can you help me understand why they were rendered the way they were? I took a little trip through the git history, but nothing stood out for me. 

Also, before we merge this, can we get a screenshot of the before and after for the example page from the issue?

gumaerc:
@pdpinch There are screenshots in the `hugo-course-publisher` PR linked above.  Basically, the story with this is that the "template" OCW seems to be using for tables (denoted by `BEGIN TABLE HEADER (for MIT OCW Table Template 2.51)`) includes rows that use the `colspan` attribute on a cell with a `strong` tag as an inline header that breaks the flow of the table as so:

```html
<tr class="alt-row">
  <td colspan="4"><strong>Classes and Objects</strong></td>
</tr>
<tr class="row">
  <td>L 7</td>
  <td>Classes and Objects</td>
  <td>Text: 2.5&ndash;2.10, 3.1&ndash;3.8</td>
  <td>Text: 2.6&ndash;2.10, 3.1&ndash;3.8</td>
</tr>
```

Hugo's implementation of the Goldmark renderer does not implement the `colspan` attribute in any way, so we need a different way to render the table and have the inline header not be contained by a single cell.  Goldmark is based on the "CommonMark" standard, the same as Github flavored Markdown, and theoretically the way to represent colspan would be:

```md
| **Classes and Objects** | > | > | > |
```

However, either Goldmark doesn't support this or Hugo doesn't implement it.  I'm not exactly sure where it's not being brought in.  In general I've been trying to avoid rendering raw HTML in the Markdown, so previously I had broken and recreated the table and inserted the header in between.  Instead of doing that, to keep the table whole, I'm now rendering a shortcode in that first cell that renders the header in a div with the `position-absolute` class and `nbsp;` character entities in the other cells.  I've tested this in all browsers at desktop and mobile resolutions and it works as expected, and is reflected in the screenshots.

mbertrand:
Left a comment on the related PR https://github.com/mitodl/hugo-course-publisher/pull/223, noticed an issue with at least 1 course, not sure if the fault is with this PR or that one.

mbertrand (approved):
👍

mbertrand (approved):
👍

PR #86, opened 2020-09-24
change out markdown-builder for markdown-doc-builder

aliceriot:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

none

#### What's this PR do?

removes the markdown-builder dependency to get rid of the annoying inclusion of [husky](https://www.npmjs.com/package/husky)

check that the code makes sense and that the markdown produced is the same as before. to test that what I did was

1. create an empty git repo in `/tmp/md-test`
2. run `ocw-to-hugo` with the `example-courses.json` list and send the output into `/tmp/md-test`
3. commit that markdown to the repo
4. check out this branch, run `yarn install`, and then re-run `ocw-to-hugo`. the git repo should show no changed files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "cli-progress": "^3.4.0",
     "dotenv": "^8.2.0",
     "js-yaml": "^3.13.1",
-    "markdown-builder": "^0.9.0",
+    "markdown-doc-builder": "^1.3.0",
     "title-case": "^3.0.2",
     "tmp": "^0.2.1",
     "turndown": "^6.0.0",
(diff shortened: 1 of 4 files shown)


gumaerc (approved):
:+1: LTGM, although after some investigation I don't believe we are using that `generateCourseFeatures` function anymore.  A while back we moved Course Features into the Course Info section.  I believe now the data is just being inserted into the course home page as YML front matter and being built out by a partial on the `hugo-course-publisher` side.  I'll make an issue to get rid of it.

PR #91, opened 2020-09-28
Remove unneeded fs commands

### LINKED ISSUE #82: Remove path exist checks when not needed

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #82 

#### What's this PR do?
Removes some redundant fs commands for code cleanliness and performance

Convert all courses and verify that the output has not changed at all

### DIFF
--- a/src/aws_sync.js
+++ b/src/aws_sync.js
@@ -6,7 +6,7 @@ const AWS = require("aws-sdk")
 require("dotenv").config()
 const cliProgress = require("cli-progress")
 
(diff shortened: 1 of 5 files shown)


mbertrand (approved):
LGTM, worked fine on `course_json_examples/example_courses.json`

PR #99, opened 2020-10-07
Exit with non-zero exit code only on verbose, if a non-fatal error occurs

noisecapella:
#### Pre-Flight checklist

- [x] Testing

None

#### What's this PR do?
Since most hugo builds will have some non-fatal error which is logged, the exit code is almost always non-zero. Since this is unhelpful in production but helpful in debugging, we should only exit with a non-zero exit code in this case if 

Convert one of the courses with a CSS error. If `--verbose` is in your command line you should see a non-zero exit code, but if `--verbose` is not there you should not.

### DIFF
--- a/src/bin/index.js
+++ b/src/bin/index.js
@@ -80,7 +80,7 @@ run()
           "Found errors which were logged: ",
           loggers.memoryTransport.logs
         )
+        process.exit(1)
       }
-      process.exit(1)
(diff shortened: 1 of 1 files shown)


gumaerc (approved):
:+1: LGTM

markbreedlove:
👍

PR #100, opened 2020-10-07
Error if master JSON is missing for a course and the user has a course list

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
 - Right now missing master JSON files cause a course to be skipped. The error is logged but it's not a fatal error. This changes that behavior to make it a fatal error if the user has specified a course list in the command line.
 - options are now set to `helpers.runOptions` and not passed into `scanCourses`
 - Courses which are present in the courses JSON but are missing directories will result in an error, if a courses JSON is set

Remove or rename the master JSON file for a course, then convert it. You should get an error when the processing reaches that course.

### DIFF
--- a/src/file_operations.js
+++ b/src/file_operations.js
@@ -31,7 +31,7 @@ const writeBoilerplate = async outputPath => {
   }
 }
 
-const scanCourses = async (inputPath, outputPath, options = {}) => {
+const scanCourses = async (inputPath, outputPath) => {
   /*
     This function scans the input directory for course folders
   */
@@ -43,9 +43,7 @@ const scanCourses = async (inputPath, outputPath, options = {}) => {
     throw new Error("Invalid output directory")
   }
 
-  const jsonPath = options.courses
-  helpers.runOptions.strips3 = options.strips3
-  helpers.runOptions.staticPrefix = options.staticPrefix
+  const jsonPath = helpers.runOptions.courses
   const courseList = jsonPath
     ? JSON.parse(await fsPromises.readFile(jsonPath))["courses"]
     : (await fsPromises.readdir(inputPath)).filter(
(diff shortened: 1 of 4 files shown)


src/file_operations.js
```diff
@@ -60,22 +58,28 @@ const scanCourses = async (inputPath, outputPath, options = {}) => {
...
-  } else {
+  if (numCourses === 0) {
     console.log(NO_COURSES_FOUND_MESSAGE)
+    return
+  }
+
+  // populate the course uid mapping
+  const courseUidsLookup = {}
+  for (const course of courseList) {
+    if (!(await directoryExists(path.join(inputPath, course)))) {
```
gumaerc, line 69: When I ran this against my `aws sync`'d local copy of `open-learning-course-data`, I got this error:

```
gumaerc@gumaerc-laptop:~/Code/ocw-to-hugo$ node . -i ../hugo-course-publisher/private/courses -o ../hugo-course-publisher-2/site/content
Error: Error: Missing course directory for edx_courses.json
    at scanCourses (/home/gumaerc/Code/ocw-to-hugo/src/file_operations.js:70:13)
    at async run (/home/gumaerc/Code/ocw-to-hugo/src/bin/index.js:66:3)
```
It seems that `directoryExists` is returning true even if it's a file and not a directory.  While iterating `courseList`, we should skip entries that are not directories.

gumaerc (changes requested):
Just had one slight issue:

src/file_operations.js
gumaerc: I should note though that it ran fine after removing the stray json files from that directory, although it should ignore them.  Also specifying courses that didn't exist in a `courses.json` file errored as expected.

noisecapella: Good catch, should be fixed now. I added a file to `test_data/courses` to make sure the tests detect this issue in the future

gumaerc (approved):
LGTM :+1:

PR #102, opened 2020-10-15
Add topicslist for easier rendering in hugo

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of https://github.com/mitodl/hugo-course-publisher/issues/240

#### What's this PR do?
Adds a `topicslist` field to the frontmatter so that a list of topics can be easily generated in the template. The `topics` dictionary isn't easily traversed in hugo without some complicated template logic.

Convert a course and check the output of a course markdown file to see that `topicslist` matches the

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -125,6 +125,13 @@ const getConsolidatedTopics = courseCollections => {
   })
   return topics
 }
+
+const getConsolidatedTopicsList = courseCollections =>
+  _.uniq(
+    courseCollections
+      .map(collection => collection.ocw_subfeature || collection.ocw_feature)
+      .filter(Boolean)
+  )
 /* eslint-disable camelcase */
 
 const getYoutubeEmbedCode = media => {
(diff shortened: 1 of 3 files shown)


gumaerc (changes requested):
I'm still not sure this is the way to accomplish this.  It seems like for the topics list, we basically just want to list the subtopics if they exist, and the topic if they don't.  I was able to accomplish this using this as a partial in `hugo-course-publisher`:

topic_list.html
```
{{ $topicsList := slice }}
{{ range $topic, $subtopics := . }}
{{ if not $subtopics }}
    {{ $topicsList = $topicsList | append (title $topic) }}
{{ else }}
    {{ range $subtopic, $specialities := $subtopics }}
    {{ $topicsList = $topicsList | append (title $subtopic) }}
    {{ end }}
{{ end }}
{{ end }}
{{ delimit $topicsList ", " }}
```

You can then call it using:

```
{{ partial "topic_list.html" $courseInfo.topics }}
```

I think it'd be best to avoid adding duplicate data if we can.  When we put a CMS in front of this in the future it might not be great to have to manage both properties.

PR #103, opened 2020-10-21
Handle an empty string for instructors

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
Handles an empty string for instructors in the master JSON. This should be fixed in https://github.com/mitodl/ocw-data-parser/pull/64 but it will take a little while to run that on all courses so this is here as a quicker temporary fix. At some point after we can revert this fix.

Set `instructors: ""` in your master JSON or use a file which already has this value, then convert the courses.


gumaerc (approved):
:+1:

PR #105, opened 2020-10-21
Remove logged error if the course is skipped and no course JSON is set

noisecapella:
#### Pre-Flight checklist

- [x] Testing

None

#### What's this PR do?
Removes a logging statement which shouldn't be there

Remove a master JSON file from a course directory and convert it without specifying a JSON file, with the `--verbose` flag. You should see no error but the output directory will not have that course in it.


gumaerc (approved):
:+1: LGTM

PR #118, opened 2020-11-12
Remove title from boilerplate

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Part of https://github.com/mitodl/hugo-course-publisher/issues/328

#### What's this PR do?
Removes "Hugo Course Publisher" from the title so it doesn't show up. hugo-course-publisher will interpret this to use its own title from the markdown parameters, or else a more sensible default.

Convert a course, then check `_index.md` at the root level and verify that it has an empty title.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -16,7 +16,7 @@ module.exports = {
       path:    "",
       name:    "_index.md",
       content: {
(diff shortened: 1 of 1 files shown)


gumaerc (approved):
:+1: LGTM

PR #119, opened 2020-11-12
Change topics to return a list with dicts

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
The topics and subtopics are only available in lowercase because hugo converts dict keys to lowercase values. We are using `title` as a workaround, but this doesn't work consistently with search because the search requires case sensitive topic values.

Only the topics should change in converted courses. This will be a breaking change but we can update the corresponding code on hugo-course-publisher so we have a seamless update.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -104,25 +104,45 @@ const getCourseSectionFromFeatureUrl = courseFeature => {
 
 /* eslint-disable camelcase */
 const getConsolidatedTopics = courseCollections => {
-  let topics = {}
-  courseCollections.forEach(courseCollection => {
-    const { ocw_feature, ocw_subfeature, ocw_speciality } = courseCollection
+  const topics = []
+  const topicsLookup = {}
+  const subtopicsLookup = {}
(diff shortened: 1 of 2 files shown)


gumaerc:
This is an interesting way to have accomplished this, but if the keys are a problem and we really need to present the data in the structure under `topics_list`, I wonder if it might be best to just adjust the `topics.html` partial to use the format of `topics_list`.  That way we could just call it `topics` and not have the same data defined twice in every course.  The `topics.html` partial would probably be a bit more verbose, but I think it could be done.  What do you think?

noisecapella:
Sure, I'll change this to use `topics` and remove the old format. We can do a release after and use that release in the hugo PR which updates `topics.html`

gumaerc (approved):
:+1: LGTM

I created an issue in `hugo-course-publisher` as a reminder to update `topics.html` during the next `ocw-to-hugo` upgrade PR: https://github.com/mitodl/hugo-course-publisher/issues/345

PR #120, opened 2020-11-13
Add course_title and course_info to section pages

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Part of https://github.com/mitodl/hugo-course-publisher/issues/328

#### What's this PR do?
Adds course info and course title to section front matter.

#### How should this be manually tested?
You should see the new fields in the output markdown for the course

#### Any background context you want to provide?
I think this competes with https://github.com/mitodl/hugo-course-publisher/issues/326 but I'm not sure which path we're choosing for this

PR #128, opened 2020-11-20
Remove course_info from section pages

noisecapella:
Reverts a part of mitodl/ocw-to-hugo#120. It removes `course_info` from course section pages to try to decrease hugo memory use


pdpinch:
Let's hold off on reviewing and merging this until after 11/23.

Issue #129, opened 2020-11-23
ocw-to-hugo should be able to clear the destination directory before conversion


gumaerc:
As a developer, I'd like to be able to run `ocw-to-hugo` and be sure that there is no old markdown data left behind in the destination directory that could cause issues with the build.

#### Acceptance Criteria:

- [ ] An optional argument can be passed to `ocw-to-hugo` that triggers clearing of the destination directory

PR #131, opened 2020-11-23
add optional argument to clear destination directory before conversion

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/129

#### What's this PR do?
This PR adds the `--rm` argument to `ocw-to-hugo`.  If used, the output directory (the directory passed with the `-o` argument) will have its contents recursively removed before the conversion run.

Create a directory with a test file in it, then run `ocw-to-hugo` targeting that directory with `-o` and also pass the `--rm` argument.  After the run is complete, ensure that the test file is gone.


mbertrand (approved):
👍

PR #133, opened 2020-11-25
upgrade turndown

aliceriot:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

updates turndown, which was recently updated to no longer depend on JSDom, instead using [domino](https://github.com/fgnass/domino), which is much faster. ocw-to-hugo now runs in around half the time it used to.

I ran 
```
time node ./src/bin/index.js -c course_json_examples/example_courses.json -i private/courses -o private/output --strips3`
```

to test. on master:

```
62.59s user 3.83s system 127% cpu 52.244 total
```

with turndown v7.0.1:

```
25.27s user 2.89s system 125% cpu 22.380 total
```

we need to ensure that this change doesn't introduce any problems into ocw-to-hugo's output. I have run the example courses through and those render in hugo-course-publisher without any errors or (apparent) issues. I also diffed the output of those courses with the output on ocw-to-hugo master and there are a few differences but nothing that I think will be problematic.

but as far as testing this I would check that!

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -204,19 +204,6 @@
     lodash.unescape "4.0.1"
     semver "5.5.0"
 
-abab@^2.0.3:
-  version "2.0.3"
-  resolved "https://registry.yarnpkg.com/abab/-/abab-2.0.3.tgz#623e2075e02eb2d3f2475e49f99c91846467907a"
-  integrity sha512-tsFzPpcttalNjFBCFMqsKYQcWxxen1pgJR56by//QwvJc4/OUS3kPOOttx2tSIfjsylB0pYu7f5D3K1RCxUnUg==
-
-acorn-globals@^6.0.0:
-  version "6.0.0"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1629,43 +1405,6 @@ js-yaml@3.13.1, js-yaml@^3.13.0, js-yaml@^3.13.1:
     argparse "^1.0.7"
     esprima "^4.0.0"
 
-jsbn@~0.1.0:
-  version "0.1.1"
-  resolved "https://registry.yarnpkg.com/jsbn/-/jsbn-0.1.1.tgz#a5e654c2e5a2deb5f201d96cefbca80c0ef2f513"
-  integrity sha1-peZUwuWi3rXyAdls77yoDA7y9RM=
-
-jsdom@^16.2.0:
```
noisecapella, line 1637: :sunglasses:

noisecapella (approved):
LGTM :+1:

PR #143, opened 2020-12-03
[draft] implement caching of converted markdown, data between builds

alicewriteswrongs:
draft implementation of caching. basically this saves converted courses as gzipped tar archives and copies the course data JSON files to the cache directory (`~/.cache/ocw-to-hugo` for now). then when running a conversion it checks to see if the input data is older than the cached archive. If the input data is older than the archive it inflates the archive in the output directory instead of running the conversion. This saves us from having to run the expensive html -> markdown conversion for courses which haven't changed since the last time the program has been run.

note that, as a draft implementation, things are still messy:

- I freely mixed async and sync `fs` methods
- no tests have been written
- options like `--no-cache` (to turn the functionality off altogether) and an option for setting the cache location aren't set up

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# ocw-to-hugo's cource cache
+.cache
+
 # C extensions
 *.so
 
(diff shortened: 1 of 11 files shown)


alicewriteswrongs:
I think given that we're now out of the ocw-to-hugo age and unlikely to be running it again much in the future we can probably just close this. If we need more speed in the future this can always be reopened.

PR #144, opened 2020-12-03
RFC: add caching of converted course content

alicewriteswrongs:

We didn't have RFCs here yet, so I just put them in the same directory setup we use in open-discussions.

#### What's this PR do?

RFC for adding caching support to `ocw-to-hugo`.

Rendered: https://github.com/mitodl/ocw-to-hugo/blob/bbd895112d9e0629a61ce4a3b7d784200ef7b67a/docs/rfcs/Caching.md


gumaerc:
I went into reading this with a couple questions in my head and basically all of them were answered by the time I was done reading.  I think this is a simple and effective way to save on build time.

One thing I'm wondering though is about the age comparison is using modified time vs created time.  I'm not sure if it would ever be an issue, but I can think of something like moving the data directory on the server or updating permissions on the files causing all the modified dates to change, not reflecting when the file was actually created.  So, maybe it would be better to use the creation date?  That's the only question I came out of this with.

aliceriot:
hmm that's true and something to think more about. Right now in the draft implementation I'm using `fstatSync(filename).mtime` to get the modified time, and I wrote a little function to find the most-recent modified time for a file in the directory. So that would I think lead to an unnecessary re-conversion (i.e. the cache would be busted needlessly) if permissions were changed on a file. I don't think moving the overall course directory would do that though - the `mtime` for all the files inside the course directories wouldn't change just because their parent directories changed, assuming that they were moved with `mv` - I think if `rsync` or something that actually writes new files were used then everything would like 'new' as far as `mtime` is concerned.

To use the create time, are you suggesting using the most-recent create time on a file to capture when the course directory was last written to? I think that could work - but I'm not exactly sure when that would be updated.

From the command line, if I do:

```
echo beep > foo.txt
```

and then wait a minute and run

```
echo bip > foo.txt
```

then it looks like `fstatSync("foo.txt").ctime` (created at time) does reflect when the second command was run (as does `.mtime`, since I think they are equivalent here). I think my POSIX-fu is not high enough to know off-hand when these values can be different, haha.

aliceriot:
it seems that ctime will reflect changes in permissions, but mtime won't, and should only change when the file contents are actually changed: https://www.unixtutorial.org/atime-ctime-mtime-in-unix-filesystems/

gumaerc:
Yea, I'm not super familiar with details like this either, but it does seem like `mtime` will work.  In theory, we might not need to check the `mtime` of all the static resources in the directory either and could get away with just checking the `_parsed.json` files because when new static assets are added they should be listed in there under `course_files` and hence the json file should have a new `mtime`.

aliceriot:
Updated to reflect some recent conversations

alicewriteswrongs:
I closed the PR, going to close the issue too. If we ever need this in the future, there is a draft implementation of basic caching here: #143

Issue #154, opened 2021-01-05
uid not preserved on course home pages


gumaerc:
### Steps to Reproduce

Convert any number of courses

### Expected Behavior

The course home page of each course has a `uid` front matter property with the course home page's original UID from Plone, and the menu identifier matches this uid

### Actual Behavior

The course home page does not have a `uid` property and the menu identifier is `course-home`

PR #155, opened 2021-01-05
preserve UID on course home pages when possible

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/154

#### What's this PR do?
This PR sets the `uid` and menu `identifier` property on course home pages to the original uid from Plone

Convert any number of courses and assert that each course home page markdown file contains a `uid` property matching the uid of the course home page from the source data.


noisecapella:
Is this ready for review yet?

gumaerc:
@noisecapella yes, ready for a look now.

src/markdown_generators.js
```diff
@@ -226,6 +228,7 @@ const generateCourseSectionFrontMatter = (
     Generate the front matter metadata for a course section given a title and menu index
     */
   const courseSectionFrontMatter = {
+    uid:       pageId,
```
noisecapella, line 231: This is value is `course_home` for `res-2-006-girls-who-build-cameras-summer-2016`. Is that the right value for this?

gumaerc: I'll need to look into this tomorrow, but I think there might be a data issue with some of the courses.  That value is a default in the event the course home page is not found for whatever reason.  Stay tuned.

mbertrand:
I just ran this branch against course `18-01-single-variable-calculus-fall-2006` and the page that previously had no uid (`18-01-single-variable-calculus-fall-2006/sections/readings/course-reader.md`) now does!

```
---
uid: 98fdab81f6f5a25624a5128fe1261e9b
title: Course Reader
course_id: 18-01-single-variable-calculus-fall-2006
type: course
layout: course_section
---
```

src/markdown_generators.js
gumaerc: @noisecapella I spent some time looking into this today, and digging through the Plone JSON exports for that course I didn't find any pages of the type `CourseHomeSection`.  It doesn't seem to have a "course home" in the original data, in which case the default of `course_home` makes sense for now I think.

gumaerc: Actually, I just found that for "resource" courses like this, the type seems to be `SRHomePage`.  This type will have to be added to the types parsed by `ocw-data-parser` and added to the `course_pages` array.  We can either merge this as-is for now, or hold off until I can push through an `ocw-data-parser` PR to pull in those other page types.

gumaerc: @noisecapella PR here for that: https://github.com/mitodl/ocw-data-parser/pull/117

gumaerc:
@noisecapella this could use another look

noisecapella:
Can we leave the `uid` blank instead of having a default of `course-home`?

gumaerc:
@noisecapella all set, but @mbertrand should be aware of this

gumaerc:
Whoops, a bit of a slip up there and almost committed the wrong type match for SRHomePage but that's all set now.

noisecapella (approved):

PR #157, opened 2021-01-07
Change PDF extension handling to be case insensitive

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Fixes #156 

#### What's this PR do?
Updates `.pdf` string replacement to be case insensitive, and also refactors `resolveUids` to separate the match replacement code from the link resolution code. 

#### How should this be manually tested?
Convert the example courses on master and on this PR. Most content should be identical, but you should see these differences in `16-01-unified-engineering-i-ii-iii-iv-fall-2005-spring-2006/sections/thermo-propulsion`:
 - Some links in `_index.md` and `thermodoclist.md`, for example  `.../thermo-propulsion/p4.PDF` should now just be `.../thermo-propulsion/p4`
 - Files that used to look like `p2.PDF.md` should now be `p2.md`, but they are identical

PR #158, opened 2021-01-07
Refactor resolveRelativeLinks

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

None

#### What's this PR do?
I did some refactoring of `resolveRelativeLinks` during my work on #157  to make it easier to make changes in the future, and it looks like there are some bug fixes too.

Convert the example courses using the master branch and this one, and compare the results. You should see mostly identical files, except:
 - course `20-020-introduction-to-biological-engineering-design-spring-2009` has a few markdown files where a `GETPAGESHORTCODE` instance used to be added to certain image URLs. I think that was a bug where the prefix of the image URL matched a course page URL. I think at the moment we aren't doing any processing on image URLs, just regular links.
 - `6-042j-mathematics-for-computer-science-spring-2015` has one file `vertical-dcc88d262213.md` which has a similar image link issue, but there is also a link which is being processed correctly where it didn't use to be. To test this, go to the [production](https://ocwnext.odl.mit.edu/courses/6-042j-mathematics-for-computer-science-spring-2015/sections/probability/tp11-2/vertical-dcc88d262213/) page and scroll to the bottom. There should be a link called "BackIntro to Discrete Probability", but the link is broken. If you look at the same link on your local instance it should work correctly.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -276,6 +276,79 @@ const resolveUids = (htmlStr, page, courseData, courseUidsLookup) => {
   return htmlStr
 }
 
+const resolveRelativeLink = (url, courseData) => {
+  // ensure that this is not resolveuid or an external link
+  if (!url.includes("resolveuid") && url[0] === "/") {
+    // split the url into its parts
+    const parts = url.split("/").filter(part => part !== "")
+    /**
+     * disassembles the OCW URL based on the following patten:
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
👍

PR #169, opened 2021-01-20
add instructor_insights layout

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to https://github.com/mitodl/ocw-course-hugo-theme/issues/9

#### What's this PR do?
This PR sets the `layout` front matter value on Instructor Insights pages to `instructor_insights` so that a layout may be defined specifically for these pages in `ocw-course-hugo-theme`.

Convert any course that contains an Instructor Insights section (for example, `6-034-artificial-intelligence-fall-2010`) and ensure that the markdown representing that section and any children has the front matter `layout` property set to `instructor_insights`.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -78,8 +78,8 @@ const generateMarkdownRecursive = (page, courseUidsLookup, courseData) => {
   const parent = hasParent ? parents[0] : null
   const inRootNav = page["parent_uid"] === courseData["uid"]
   const isInstructorInsightsSection =
-    page["short_url"] === "instructor-insights" ||
-    (hasParent && parent["short_url"] === "instructor-insights")
+    page["type"] === "ThisCourseAtMITSection" ||
+    (hasParent && parent["type"] === "ThisCourseAtMITSection")
   const layout = isInstructorInsightsSection
     ? "instructor_insights"
     : "course_section"
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -79,7 +79,9 @@ const generateMarkdownRecursive = (page, courseUidsLookup, courseData) => {
   const inRootNav = page["parent_uid"] === courseData["uid"]
   const isInstructorInsightsSection =
     page["type"] === "ThisCourseAtMITSection" ||
-    (hasParent && parent["type"] === "ThisCourseAtMITSection")
+    page["short_url"] === "instructor-insights" ||
+    (hasParent && parent["type"] === "ThisCourseAtMITSection") ||
+    (hasParent && parent["short_url"] === "instructor-insights")
   const layout = isInstructorInsightsSection
     ? "instructor_insights"
     : "course_section"
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -77,8 +77,15 @@ const generateMarkdownRecursive = (page, courseUidsLookup, courseData) => {
   const hasParent = parents.length > 0
   const parent = hasParent ? parents[0] : null
   const inRootNav = page["parent_uid"] === courseData["uid"]
+  const isInstructorInsightsSection =
+    page["short_url"] === "instructor-insights" ||
+    (hasParent && parent["short_url"] === "instructor-insights")
+  const layout = isInstructorInsightsSection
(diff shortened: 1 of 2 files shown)


src/markdown_generators.js
```diff
@@ -77,8 +77,15 @@ const generateMarkdownRecursive = (page, courseUidsLookup, courseData) => {
   const hasParent = parents.length > 0
   const parent = hasParent ? parents[0] : null
   const inRootNav = page["parent_uid"] === courseData["uid"]
+  const isInstructorInsightsSection =
+    page["short_url"] === "instructor-insights" ||
+    (hasParent && parent["short_url"] === "instructor-insights")
+  const layout = isInstructorInsightsSection
+    ? "instructor_insights"
+    : "course_section"
```
aliceriot, line 85: would it be relatively straightforward to add a test to make sure `"instructor_insights"` shows up when we expect it to?

aliceriot (changes requested):
just one question really

src/markdown_generators.js
gumaerc: yep, working on that now

gumaerc:
@aliceriot test added

src/markdown_generators.js
```diff
@@ -77,8 +77,15 @@ const generateMarkdownRecursive = (page, courseUidsLookup, courseData) => {
   const hasParent = parents.length > 0
   const parent = hasParent ? parents[0] : null
   const inRootNav = page["parent_uid"] === courseData["uid"]
+  const isInstructorInsightsSection =
+    page["short_url"] === "instructor-insights" ||
```
pdpinch, line 81: Does this mean that only a section with the a `short_url` of `instructor-insights` will be process this way? Does this restrict us to never change the page title? 

I think it would be better future-proofed to identify this as a page type instead of by URL.

gumaerc: @pdpinch This is a good point and I thought about this.  I had thought that Instructor Insights pages just had a `type` of `CourseSection` (their child pages do) but upon further inspection, the root Instructor Insights page seems to have a `type` of `ThisCourseAtMITSection`.  Does that seem right to you?

gumaerc:
@aliceriot Could you take another look at this?  I'd like to get it merged to rebase another branch I'm working on

aliceriot (approved):
As long as we're relatively confident that the `"ThisCourseAtMITSection"` change will get them all then it looks good to me!

gumaerc:
image
image

As you can see, there does seem to be a slight discrepancy searching for `"type": "ThisCourseAtMITSection"` vs `"short_url": "instructor-insights"`.  @pdpinch What do you think, should we allow either condition to trigger setting the `instructor_insights` layout?

gumaerc:
It seems the following courses have an Instructor Insights section where the `type` is just set to `CourseSection`:

```
11-124-introduction-to-education-looking-forward-and-looking-back-on-education-fall-2011
11-131-educational-theory-and-practice-iii-spring-2012
15-279-management-communication-for-undergraduates-fall-2012
18-330-introduction-to-numerical-analysis-spring-2012
18-904-seminar-in-topology-spring-2011
18-s996-category-theory-for-scientists-spring-2013
21w-732-science-writing-and-new-media-fall-2010
22-033-nuclear-systems-design-project-fall-2011
4-241j-theory-of-city-form-spring-2013
6-685-electric-machines-fall-2013
7-346-virus-host-interactions-in-infectious-diseases-spring-2013
8-592j-statistical-physics-in-biology-spring-2011
9-00sc-introduction-to-psychology-fall-2011
es-s10-drugs-and-the-brain-spring-2013
```

src/markdown_generators.js
pdpinch: That's correct. The content type was originally called "This Course at MIT" but then was renamed in the CMS, but not in the code.

pdpinch:
Fixing this in upstream in Plone will take some time, if it's even possible. If it's not too much trouble, I think the best choice for now is to 

> match either on the ThisCourseAtMITSection type or a short_url proprety of instructor-insights

PR #170, opened 2021-01-21
add quote shortcode

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to https://github.com/mitodl/ocw-course-hugo-theme/issues/10

#### What's this PR do?
This PR catches a pattern used in Instructor Insights pages that looks like this:

```html
<div class="pullquote right">
  <p class="quote">I think stories are an important element of education, and if you strip them out, you don't have
    much left that can possibly be inspiring.</p>
  <p class="sig">&mdash; Patrick Winston</p>
</div>
```

and turns them into this:

```md
{{< quote "I think stories are an important element of education, and if you strip them out, you don't have much left that can possibly be inspiring." "— Patrick Winston" >}}
```

This will allow us to implement a custom shortcode in `ocw-course-hugo-theme` to layout / style the information how we want.

Convert `6-034-artificial-intelligence-fall-2010` and inspect the output.  Look specifically at the `teaching-heuristics` page and ensure that the quote on that page has been rendered as a shortcode as shown above.

### DIFF
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -216,7 +216,7 @@ turndownService.addRule("getpageshortcode", {
     return false
   },
   replacement: (content, node, options) => {
-    const children = Array.prototype.slice.call(node.childNodes)
+    const children = Array.from(node.childNodes)
     if (!children.filter(child => child.nodeName === "IMG").length > 0) {
       // if this link doesn't contain an image, escape the content
       // except first make sure there are no pre-escaped square brackets
@@ -251,7 +251,7 @@ turndownService.addRule("quoteshortcode", {
   },
   replacement: (content, node, options) => {
     try {
-      const children = Array.prototype.slice.call(node.childNodes)
+      const children = Array.from(node.childNodes)
       const quoteP = children.find(
         child =>
           child.nodeName === "P" && child.getAttribute("class") === "quote"
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -259,8 +259,8 @@ turndownService.addRule("quoteshortcode", {
       const sigP = children.find(
         child => child.nodeName === "P" && child.getAttribute("class") === "sig"
       )
-      const quote = quoteP.textContent
-      const sig = sigP.textContent
+      const quote = helpers.escapeDoubleQuotes(quoteP.textContent)
+      const sig = helpers.escapeDoubleQuotes(sigP.textContent)
       return `{{< quote "${quote}" "${sig}" >}}`
     } catch (err) {
       loggers.fileLogger.error(err)
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -238,6 +238,37 @@ turndownService.addRule("getpageshortcode", {
 })
 
 /**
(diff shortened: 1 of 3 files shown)


src/turndown.js
```diff
@@ -237,6 +237,37 @@ turndownService.addRule("getpageshortcode", {
...
+  filter: (node, options) => {
+    if (node.nodeName === "DIV" && node.getAttribute("class")) {
+      if (node.getAttribute("class").includes("pullquote")) {
+        return true
+      }
+    }
+    return false
+  },
+  replacement: (content, node, options) => {
+    try {
```
noisecapella, line 253: Is there any error in particular we're trying to capture? It looks like many of the other rules `replacement` don't have a similar try/catch block

src/turndown.js
```diff
@@ -237,6 +237,37 @@ turndownService.addRule("getpageshortcode", {
...
+      const children = Array.prototype.slice.call(node.childNodes)
```
noisecapella, line 254: This can be `Array.from(node.childNodes)`

src/turndown.js
```diff
@@ -237,6 +237,37 @@ turndownService.addRule("getpageshortcode", {
...
+      const quoteP = children.find(
+        child =>
+          child.nodeName === "P" && child.getAttribute("class") === "quote"
+      )
+      const sigP = children.find(
+        child => child.nodeName === "P" && child.getAttribute("class") === "sig"
+      )
+      const quote = quoteP.textContent
+      const sig = sigP.textContent
+      return `{{< quote "${quote}" "${sig}" >}}`
```
noisecapella, line 264: Do we need to do any escaping of quote characters which might be in `textContent`?

noisecapella (changes requested):

src/turndown.js
gumaerc: So I had this in there to debug the output while developing this, but I think there is value in having a generic try catch on turndown replacement functions.  Any uncaught errors from running HTML through turndown do not produce an error in the calling code.  I just tested trying to manually throw an exception with `throw` in a turndown rule and it was swallowed.  We could ask the author about why it behaves that way, but I have a feeling that due to the chaotic nature of processing raw HTML into markdown errors are suppressed.  A generic try catch that logs the error into a file seems like a good workaround for now.  What do you think?

src/turndown.js
gumaerc: That's probably a good idea, although I have been testing converting all courses and I don't believe there have been any issues.

gumaerc:
@noisecapella back to you with a question about error handling above.  Maybe that's something we can address in a separate PR?

gumaerc:
@noisecapella I just had another thought: now that we aren't including `ocw-to-hugo` as an npm dependency in the build pipeline, we should probably hold off on merging this until the shortcode itself has been created in the theme and released.  Otherwise, we will get build errors about it not existing.

gumaerc:
This is blocked by [this](https://github.com/mitodl/ocw-course-hugo-theme/pull/22) PR

PR #175, opened 2021-01-28
Update and fix link processing

### LINKED ISSUE #159: Some links have `courses` prefixed when they shouldn't
See this comment https://github.com/mitodl/ocw-to-hugo/pull/157#issuecomment-756428583

### Steps to Reproduce

Go to `16-01-unified-engineering-i-ii-iii-iv-fall-2005-spring-2006/sections/thermo-propulsion` and go to the `thermo-propulsion` section. Links should be relative to the course itself and not based off of `/courses`.

### LINKED ISSUE #173: PDF's pages not linked properly in 18-01-single-variable-calculus-fall-2006
### Steps to Reproduce

Visit https://ocwnext.odl.mit.edu/courses/18-01-single-variable-calculus-fall-2006/sections/readings/course-reader/

### Expected Behavior

PDF links actually link to PDF pages

### Actual Behavior

They do not

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #159 
Fixes #173 

#### What's this PR do?

 - Removes use of `getpage` shortcode and uses relative links instead. All links should be functional to my knowledge, let me know if you see an edge case where a link doesn't look right.
 - `pathToChildRecursive` and `getHugoPathSuffix` are removed. Instead, paths are generated in the beginning all at once in `buildPaths` and `buildPathRecursive`.

Most of the changed lines are whitespace due to indentation changes in tests, so you may want to add a `?w=1` to remove whitespace: https://github.com/mitodl/ocw-to-hugo/pull/175/files?w=1

Convert all courses on master, then convert all courses with this branch, then compare the two directories with a file comparison tool. Use `--verbose` to show any errors logged in the output. There will be many differences but in general you should see `{{ getpage ... }}` shortcodes converted to simple relative links, and those links should start `/section` instead of `/course/course-slug-goes-here/section/...`. You should be able to convert a course

Also, check that the issues are fixed:
 - #159 - PDF links should work on `16-01-unified-engineering-i-ii-iii-iv-fall-2005-spring-2006` at `/sections/thermo-propulsion`
 -  #173 - PDF links should work on `18-01-single-variable-calculus-fall-2006 ` at `/sections/readings/course-reader/`

Known issues:
 - There are many warnings about a missing uid `42b179986219b2af6276e22208e1479f` in `6-050j-information-and-entropy-spring-2008` because it's used as a `parent_uid` but there is no other reference. This appears in the plone JSON files too so I don't see what we can do here.
 - There are a bunch of other warnings about missing uids which could be fixed in ocw-data-parser. I made a couple issues: https://github.com/mitodl/ocw-data-parser/issues/123 and https://github.com/mitodl/ocw-data-parser/issues/124

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -178,7 +178,7 @@ const buildPathRecursive = (item, itemsLookup, courseUid, pathLookup) => {
   if (courseUid === parentUid) {
     // course is the parent, so link should be off of /sections
     const filename = page[filenameKey]
-    pathLookup[uid] = path.join(runOptions.linkPrefix, "sections", filename)
+    pathLookup[uid] = path.join("/sections", filename)
     return
   }
 
@@ -369,7 +369,7 @@ const resolveRelativeLink = (url, courseData) => {
         page = parts.slice(parts.length - 1, parts.length)[0]
       }
       // build the base of the Hugo url
-      const basePieces = [runOptions.linkPrefix, "sections", ...sections]
+      const basePieces = ["/sections", ...sections]
       if (page.includes(".") && !page.includes(".htm")) {
         // page has a file extension and isn't HTML
         for (const media of courseData["course_files"]) {
@@ -490,8 +490,15 @@ const stripSuffix = suffix => text => {
   }
   return text
 }
-
 const stripPdfSuffix = stripSuffix(".pdf")
+
+const stripPrefix = prefix => text => {
+  if (text.toLowerCase().startsWith(prefix.toLowerCase())) {
+    return text.slice(prefix.length)
+  }
+  return text
+}
+const stripSlashPrefix = stripPrefix("/")
 
 const replaceSubstring = (text, index, length, substring) =>
   `${text.substring(0, index)}${substring}${text.substring(index + length)}`
@@ -519,6 +526,7 @@ module.exports = {
   isCoursePublished,
   runOptions,
   stripPdfSuffix,
+  stripSlashPrefix,
   replaceSubstring,
   applyReplacements,
   getPathFragments,
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -167,14 +167,17 @@ const buildPathRecursive = (item, itemsLookup, courseUid, pathLookup) => {
   }
 
   if (!parentItem) {
-    loggers.fileLogger.error(`Missing parent ${parentUid}`)
+    loggers.fileLogger.error(`Missing parent ${parentUid}, parent of ${uid}`)
     return
   }
 
   if (!pathLookup[parentUid]) {
     buildPathRecursive(parentItem, itemsLookup, courseUid, pathLookup)
     if (!pathLookup[parentUid]) {
-      throw new Error(`Unable to find path for ${parentUid}`)
+      loggers.fileLogger.error(
+        `Unable to find path for ${parentUid}, parent of ${uid}`
+      )
+      return
     }
   }
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -4,12 +4,7 @@ const moment = require("moment")
 
 const fsPromises = require("./fsPromises")
 const DEPARTMENTS_JSON = require("./departments.json")
-const {
-  AWS_REGEX,
-  GETPAGESHORTCODESTART,
-  GETPAGESHORTCODEEND,
-  INPUT_COURSE_DATE_FORMAT
-} = require("./constants")
+const { AWS_REGEX, INPUT_COURSE_DATE_FORMAT } = require("./constants")
 const loggers = require("./loggers")
 
 const runOptions = {}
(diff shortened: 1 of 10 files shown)


src/markdown_generators.js
```diff
@@ -161,24 +181,11 @@ const generateMarkdownRecursive = (page, courseUidsLookup, courseData) => {
   }
 }
 
-const generateCourseInfo = courseData => ({
```
noisecapella, line 164: I removed this because it looked like it was unused, but I'm happy to add back if it's needed

gumaerc: You are correct, and I think there is other unused code that can be cleaned up as well

src/helpers.js
```diff
@@ -218,29 +260,25 @@ const resolveUidForLink = (url, courseData, courseUidsLookup, pagePath) => {
...
       // create a link to the generated PDF viewer page for this PDF file
-      const pdfPath = `${pagePath.replace("/_index.md", "/")}${
-        linkedFile["id"]
-      }`
-      return `${GETPAGESHORTCODESTART}${stripPdfSuffix(
-        pdfPath
-      )}${GETPAGESHORTCODEEND}`
+      const parent = pathLookup[parentUid]
+      if (parent) {
+        const pdfPath = path.join(parent, linkedFile["id"])
```
gumaerc, line 273: I encountered an issue here with the course `18-06sc-linear-algebra-fall-2011`.  There are a couple of sections that have links to PDF's that were not functioning.  I dug into it and the HTML was being generated for them, so I was confused at first about what was wrong, but it seems the issue has to do with case sensitivity.  The URL's are being rendered into the markdown with uppercase letters, and the markdown files themselves have uppercase letters in the filenames.  However, it seems when Hugo generates the HTML it is lowercasing the filename (`MIT18_06SCF11_ex1.md` turns into a folder named `mit18_06scf11_ex1` with an `index.html` inside) and it only works if you access the URL lower case.

gumaerc: This is an issue with the Hugo dev server but it would also most certainly be an issue with AWS hosted content as S3 / Cloudfront is most certainly case sensitive.

src/helpers.js
```diff
@@ -330,7 +369,7 @@ const resolveRelativeLink = (url, courseData) => {
         page = parts.slice(parts.length - 1, parts.length)[0]
       }
       // build the base of the Hugo url
-      const newUrlBase = path.join("courses", courseId, "sections", ...sections)
+      const basePieces = ["/sections", ...sections]
```
gumaerc, line 372: I wasn't exactly sure where to put this comment, as it has a bit wider of an implication than just resolving relative urls.  I tested this PR all the way through by generating markdown for all courses and then running it through the production build to make sure everything renders properly.  It seems that absolute URLs rendered in markdown do not respect the `baseUrl` argument / configuration value.  For example, if I use the starter to render all course sites using this command:

`./build.sh -o ~/Code/ocw-www/site/public -c ~/Code/ocw-to-hugo/private/test -b http://localhost:3000/courses`

...where `~/Code/ocw-to-hugo/private/test` is the markdown output from this PR, then I browse to the course `18-06sc-linear-algebra-fall-2011`, then to the `ax-b-and-the-four-subspaces/exam-1/` section, none of the links work because they do not respect `baseUrl` when rendered as part of `ocw-www`.  For example:

`http://localhost:3000/courses/18-06sc-linear-algebra-fall-2011/sections/ax-b-and-the-four-subspaces/exam-1/`

Then click the link for Exams (PDF) and you are brought to:

`http://localhost:3000/courses/18-06sc-linear-algebra-fall-2011/sections/ax-b-and-the-four-subspaces/exam-1/MIT18_06SCF11_ex1`

So, it seems we are not totally free of the need for some kind of special shortcode to help out here.  I think we might be able to approach it differently though, and simply create a `baseUrl` shortcode as these people have done:

https://discourse.gohugo.io/t/can-i-reference-baseurl-from-content/2686/2

As long as we prefix all absolute URLs in markdown with this, it should render the baseUrl argument before the URL, resulting in a fully qualified URL in the HTML.  I haven't tested this myself, but it could be another approach as opposed to bringing back `getpage`.

gumaerc (changes requested):
A couple issues:

src/helpers.js
noisecapella: Do you think this is serious enough to deal with in this PR or would you feel comfortable creating an issue for it? I did some investigation and it seems nontrivial to solve

src/helpers.js
noisecapella: What do you think about adding a `baseUrl` argument to ocw-to-hugo and having that construct prefixes for these URLs? This PR did that originally but I removed it because I thought it was unnecessary

src/helpers.js
gumaerc: This could be solved in another PR, but in a way this PR actually causes the issue.  The links are generated properly using the `getpage` shortcode.  It's up to you though, I created an issue here: https://github.com/mitodl/ocw-to-hugo/issues/180

src/helpers.js
gumaerc: That's a possible solution, but seems like more work than just prefixing them with something like `{{< baseurl >}}` and adding a shortcode in `ocw-course-hugo-theme` that basically just has the content `{{ .Site.BaseURL }}`.  That way, when a site is generated using this content it will respect the `baseUrl` sent into the `hugo` command and you won't have to regenerate markdown to use a different one.

noisecapella: That's a good point. I'll see about updating the PR for that

src/helpers.js
noisecapella: This PR may cause that issue but unless it affects a large number of courses I think it would be simpler to deal with it in a separate PR

gumaerc: Sounds good

noisecapella:
@gumaerc FYI I noticed there is still a use of a function which was removed, and there are no tests covering that code so we never noticed it would cause a problem: https://github.com/mitodl/ocw-to-hugo/pull/175/files#diff-44aa1560b6323a1ab258ee71c8dd86aa22c75ccdfe7aab58204e90bbc5e5d676R310

I'll fix and add a test for it. I'm confused I didn't see it earlier because it didn't seem like running through all the OCW courses caused that code to be exercised.

gumaerc:
@noisecapella I was going to hold off digging further into this PR until the above points were addressed, so I didn't see this.  I'm a little confused though, are you referring to `downloadCourses`?  I remember writing tests for the AWS functionality, and there do seem to be some basic tests for it in `aws_sync_test.js`.  The tests look for specific errors related to not passing in arguments, so I would think they should be failing...

gumaerc:
@noisecapella Actually, that link you posted doesn't seem to be showing me a specific line.  Could you clarify what function you're talking about?

noisecapella:
Sorry, should have been more clear. The `pathToChildRecursive` function was removed but it's still present in `generateCourseFeatures` in `markdown_generators.js`. There is a test for that function but it doesn't hit the line with `pathToChildRecursive`

gumaerc:
Ahh yes, I see.  Good catch!

gumaerc:
@noisecapella That call to `pathToChildRecursive` in `generateCourseFeatures` is also still building a URL with the course prefix:

```js
      if (section && matchingSections.length > 0) {
        return markdown
          .newBuilder()
          .link(
            `{{% ref "${helpers.pathToChildRecursive(
              path.join("courses", courseData["short_url"], "sections"),
              matchingSections[0],
              courseData
            )}" %}}`,
            courseFeature["ocw_feature"]
          )
          .toMarkdown()
      } else return null
```

Instead I think this should be something like:

```js
      if (section && matchingSections.length > 0) {
        return markdown
          .newBuilder()
          .link(
            `{{% ref "${helpers.pathToChildRecursive(
             "sections",
              matchingSections[0],
              courseData
            )}" %}}`,
            courseFeature["ocw_feature"]
          )
          .toMarkdown()
      } else return null
```
right?

noisecapella:
@gumaerc yes I think you're right. I think we can also remove the `ref` shortcode there and just generate a full path

gumaerc:
Give it a go, but I suspect you'll maybe need to prefix it with `{{< baseUrl >}}`.  I think my reasoning behind using `ref` here was that it already takes `baseUrl` into account when building a URL:

https://gohugo.io/content-management/shortcodes/#ref-and-relref

However, it seems like it's expecting a relative URL.  The built in `ref` shortcode uses this template function:

https://gohugo.io/functions/ref/

It says: "A path without a leading `/` is first resolved relative to the given context, then to the remainder of the site."  Since all our paths fed into that in this context will have a leading slash, it should resolve them based on the root of the site.  The benefit of asking Hugo to resolve these things during build time is partly that `baseUrl` is taken into account, but also things like the case sensitivity issue we discussed.  Hugo just writes what it knows will be the URL to the resource during render time.

noisecapella:
@gumaerc I think the baseurl problem is addressed now. Can you take a look and see if I missed anything?

Regarding `generateCourseFeatures`, I noticed it was unused except by unit tests so I just removed it

gumaerc:
@noisecapella I'll take another look first thing tomorrow

gumaerc (approved):
:+1: Looking good!

PR #176, opened 2021-01-28
Refactor link processing

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to #159 

#### What's this PR do?
Some refactoring required for #175. In particular, the change makes it so `resolveRelativeLinks` does not interfere with `resolveUids` by treating them independently.

You should see no change in the output of any course compared to master

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -185,6 +185,26 @@ const getHugoPathSuffix = (page, courseData) => {
   return isParent || hasFiles ? "/_index.md" : ""
 }
 
+const applyReplacements = (matchAndReplacements, text) => {
+  const sortedMatchAndReplacements = Array.from(matchAndReplacements)
+  // this sorts in reverse order with highest index first so that we can do replacements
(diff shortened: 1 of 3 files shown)


gumaerc (approved):
:+1:  This all looks good to me.  I ran both `master` and this branch through a full build with all courses, and ran Meld on the output.  It's worth noting maybe that there was only a difference in one file.  It's  the Calendar section in `21g-504-japanese-iv-spring-2009`.  Some of the calls to `getpage` are just absolute links now:

`master`:
`([details]({{% getpage "courses/exams#3/sections" %}})`

vs.

`gs/link-processing-and-refactoring`:
`([details](/courses/21g-504-japanese-iv-spring-2009/exams#3))`

I think this is probably fine, as this is all an effort to get rid of the `getpage` shortcode and course prefix anyway.

noisecapella:
Oh, thanks for catching that. I'll check to see how my other PR handles that case

Issue #196, opened 2021-02-12
Broken Written Rules Style Guide PDF link


noisecapella:
https://ocwnext.odl.mit.edu/courses/cms-608-game-design-spring-2014/sections/syllabus/

The link for "Written Rules Style Guide" should resolve to https://ocwnext.odl.mit.edu/courses/cms-608-game-design-spring-2014/sections/syllabus/mitcms_608s14_writtenguide/. This was working but it is a regression from #191

Issue #180, opened 2021-02-08
PDF links not working in some sections


gumaerc:
### Steps to Reproduce

Generate markdown for `18-06sc-linear-algebra-fall-2011` and spin the site up using `ocw-course-hugo-theme`.  Browse to `/courses/18-06sc-linear-algebra-fall-2011/sections/ax-b-and-the-four-subspaces/exam-1/`

### Expected Behavior

The links to the PDFs go to PDF viewer pages

### Actual Behavior

They 404

### Additional context

From the PR where this was discovered:

I encountered an issue here with the course 18-06sc-linear-algebra-fall-2011. There are a couple of sections that have links to PDF's that were not functioning. I dug into it and the HTML was being generated for them, so I was confused at first about what was wrong, but it seems the issue has to do with case sensitivity. The URL's are being rendered into the markdown with uppercase letters, and the markdown files themselves have uppercase letters in the filenames. However, it seems when Hugo generates the HTML it is lowercasing the filename (MIT18_06SCF11_ex1.md turns into a folder named mit18_06scf11_ex1 with an index.html inside) and it only works if you access the URL lower case.

noisecapella:
When working on this issue we should also verify that the fix for this means that `21m-380-music-and-technology-sound-design-spring-2016/sections/syllabus` has a valid PDF link for for course entrance questionnaire

noisecapella:
There's a PDF link that's broken here which also might get fixed by this: https://ocwnext.odl.mit.edu/courses/8-13-14-experimental-physics-i-ii-junior-lab-fall-2016-spring-2017/sections/experiments/raman-spectroscopy/ at the bottom with the text "Indistinguishable Particle Effects"

PR #201, opened 2021-02-18
lowercase pdf name in links

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/180
Fixes https://github.com/mitodl/ocw-to-hugo/issues/196

#### What's this PR do?
This PR lowercases all PDF filenames used in the construction of links to PDF viewer pages.  Since we are not using the `getpage` shortcode anymore and therefore aren't getting the Hugo-generated URL to these pages, we need to do other things to conform to the way Hugo is writing these URL's.  The first one was dealing with `baseUrl`, which we have already done.  Hugo also lowercases the text in any URLs it generates, so we need to do that if we are manually generating URLs that link to Hugo generated resources.

Render markdown for a site with PDF links like `18-06-linear-algebra-spring-2010` and click the links.  They should all work on this branch, but not with content generated from `master`.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -285,7 +285,7 @@ const makePdfLink = (thisCourseId, pathObj, pathLookup) => {
     const pdfPath = path.join(
       makeCourseUrlPrefix(course, thisCourseId),
       parent,
-      id
+      id.toLowerCase()
     )
     return stripPdfSuffix(pdfPath)
   }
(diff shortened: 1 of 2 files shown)


noisecapella (approved):
Works great!

PR #208, opened 2021-02-25
approx-students turndown rule

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to https://github.com/mitodl/ocw-course-hugo-theme/issues/9

#### What's this PR do?
This PR adds a turndown rule to catch images with a file name resembling `###-approx.png` where `###` is any number.  The images are replaced with a call to the `approx-students` shortcode.

Convert a course like `6-034-artificial-intelligence-fall-2010` that has an instructor insights section with the "approximate number of students" image and inspect the output on the instructor insighs page.  You shouldn't see any references to `300-approx.png` and it should instead be replaced with a call to a shortcode that looks like: `{{ approx-students 300 }}`.

This PR is blocked by [this](https://github.com/mitodl/ocw-course-hugo-theme/pull/37) PR merging in the theme as it contains the shortcode being rendered here.

### DIFF
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -344,6 +344,29 @@ turndownService.addRule("edu_breakdown", {
 })
 
 /**
+ * Catch "approx" images and render as HTML instead
+ */
+turndownService.addRule("approx_img", {
+  filter: (node, options) => {
+    if (
+      node.nodeName === "IMG" &&
+      node.getAttribute("src").includes("-approx.png")
+    ) {
+      return true
+    }
+    return false
+  },
+  replacement: (content, node, options) => {
+    const matches =
+      node.getAttribute("src").match(/(\/|_)[0-9]+-approx.png/g) || []
+    if (matches.length > 0) {
+      const amount = matches[0].substring(1).replace("-approx.png", "")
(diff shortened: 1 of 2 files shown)


src/turndown.js
```diff
@@ -348,6 +348,29 @@ turndownService.addRule("edu_breakdown", {
...
+    ) {
+      return true
+    }
+    return false
+  },
+  replacement: (content, node, options) => {
+    const matches =
+      node.getAttribute("src").match(/(\/|_)[0-9]+-approx.png/g) || []
+    if (matches.length > 0) {
+      const amount = matches[0].substring(1).replace("-approx.png", "")
```
noisecapella, line 368: We should use a named group to capture the part of the string that we want. For example:

    const match = node.getAttribute("src").match(/(\/|_)(?<amount>[0-9]+)-approx.png/)
    if (match) {
      const amount = match.groups.amount
      return `{{< approx-students ${amount} >}}`
    }

noisecapella (changes requested):

gumaerc:
@noisecapella all set

noisecapella (approved):

Issue #224, opened 2021-03-12
use leftnav as menu key for leftnav


gumaerc:
As a developer, I'd like to be able to have a universal id for the left navigation on my site rather than the key being the course ID.  This was required back when we were running all courses inside one Hugo build with `hugo-course-publisher`, but is not necessary anymore when the Hugo context is limited to a single course.  Referring to the left navigation as `leftnav` in the menu will allow us to more easily refer to it across different projects like `ocw-studio`.

#### Acceptance Criteria:

- [x] All left nav menu items are stored in a Hugo menu called `leftnav` and `ocw-to-hugo` outputs this structure into the generated markdown
- [x] `ocw-course-hugo-theme` is set up to handle this change
- [ ] `ocw-studio` is set up to handle this change

PR #225, opened 2021-03-12
use leftnav for menu name

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/224

#### What's this PR do?
This PR changes the left navigation menu to be keyed off of `leftnav` rather than the course ID.  Previously, when the output of this project was used by `hugo-course-publisher`, all courses were built as a single Hugo site.  This required us to key the left navigation for each course on its course id to keep the entries unique.  Since courses all run in an isolated Hugo build now, we can simply key this on `leftnav`.  This will make it more straightforward to construct a menu for any given course in `ocw-studio` because it will be stored in and referenced from a single key that describes what it is and where it goes; `leftnav`.

Refer to the readme if you have never used this project before.  Generate markdown for any amount of courses and inspect the output.  For pages that have a menu entry (Syllabus, Readings, etc.) the menu should be keyed with `leftnav` as follows:

```
menu:
  leftnav:
    identifier: 84c97d818a96ffdfd496fd27a12c2c97
    name: Calendar
    weight: 20
```

Check the related issue for other PR's in repos that use this project.  They'll need to be merged at the same time as this PR.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -248,14 +248,14 @@ const generateCourseSectionFrontMatter = (
 
   if (inRootNav || listInLeftNav) {
     courseSectionFrontMatter["menu"] = {
-      [courseId]: {
+      leftnav: {
         identifier: pageId,
         name:       shortTitle || "",
         weight:     menuIndex
       }
     }
     if (parentId) {
-      courseSectionFrontMatter["menu"][courseId]["parent"] = parentId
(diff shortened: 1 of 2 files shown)


aliceriot (approved):
tried it out and everything looks good to me!

Issue #226, opened 2021-03-16
table header rendering issues


gumaerc:
### Steps to Reproduce

Render markdown for a course with a lot of information in the header rows, like `7-341-designer-immunity-lessons-in-engineering-the-immune-system-spring-2014`

### Expected Behavior

The header rows render properly with a single cell spanning all of the columns in the table, for example:

https://ocw.mit.edu/courses/biology/7-341-designer-immunity-lessons-in-engineering-the-immune-system-spring-2014/topic-summaries-and-readings/

### Actual Behavior

If the header is not a single `strong` element and is wrapped in another tag, the content is placed in the first cell:

https://ocwnext.odl.mit.edu/courses/7-341-designer-immunity-lessons-in-engineering-the-immune-system-spring-2014/sections/topic-summaries-and-readings/

PR #227, opened 2021-03-16
better colspan handling

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/226

#### What's this PR do?
Tables in OCW courses frequently have inline "header" rows which consist of a single cell with a `colspan` equal to the total amount of columns in the table.  There's usually a single `strong` element in them, sometimes wrapped in a `p` tag.  Sometimes there is extra information following the header as well.

The current turndown rule that handles tables looks for a `**` sequence (indicating a bold string created by passing `strong` through turndown) at the beginning and end of the table cell, indicating that a single bold string is the only thing in the cell.  These matches are then replaced with the `fullwidth-cell` shortcode wrapping the title with adjacent blank cells matching the total amount of columns in the table.  This works as long as that stays true, but is not very robust and is easily broken by the other scenario I described above.  Another issue is the `fullwidth-cell` shortcode itself, which forces the text content to span multiple columns by wrapping an absolutely positioned div in a relatively positioned div.  This solution doesn't work for multiple rows of content, as when the text wraps around in the absolutely positioned div, the height of its parent is unaffected.

This PR removes the existing handling of these "title rows" from the "table" markdown rule and adds a new rule.  This rule specifically looks for `td` tags with a `colspan` property.  The output of the turndown rule is a markdown table row with the first cell being populated with a `td-colspan` shortcode.  The `colspan` value from the original table cell is passed into the shortcode, and the same amount of blank cells, minus one, are added to the end.

The new `td-colspan` shortcode that replaces `fullwidth-cell` instead renders a `td` tag with a colspan property.  This places a `td` within another `td`, which is technically not allowed.  The result in the rendered HTML in all browsers I tested was two side by side `td` elements, one empty and one with the title in it along with the `colspan` attribute.  These empty `td` elements are then hidden in `ocw-course-hugo-theme`'s styles by matching on `td:empty`.  Since intentional empty cells have an `&nbsp;` character entity inserted in them, they are not hidden.

 - Read the readme if you are unfamiliar with how to generate markdown from parsed OCW JSON
 - Generate markdown for a course with the title rows I described above, `7-341-designer-immunity-lessons-in-engineering-the-immune-system-spring-2014` is a good example
 - Ensure that the outputted markdown table has a single `td-colspan` shortcode in the first cell on each title row and the value passed in matches the amount of columns in the table

For more extensive testing, rendering the markdown as a site, see the `ocw-course-hugo-theme` PR that implements the `td-colspan` shortcode [here](https://github.com/mitodl/ocw-course-hugo-theme/pull/55)

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -11,8 +11,10 @@ module.exports = {
   AWS_REGEX: new RegExp(
     /https?:\/\/open-learning-course-data(.*)\.s3\.amazonaws.com/g
   ),
-  INPUT_COURSE_DATE_FORMAT: "YYYY/M/D H:m:s.SSS",
-  SUPPORTED_IFRAME_EMBEDS:  {
+  // eslint-disable-next-line no-irregular-whitespace
+  IRREGULAR_WHITESPACE_REGEX: new RegExp(/(\|   )+\|/g),
+  INPUT_COURSE_DATE_FORMAT:   "YYYY/M/D H:m:s.SSS",
+  SUPPORTED_IFRAME_EMBEDS:    {
     "player.simplecast.com": {
       hugoShortcode: "simplecast",
       getID:         url => url.pathname.replace("/", "")
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -55,6 +55,9 @@ turndownService.addRule("table", {
   }
 })
 
+/**
+ * Catch cells with a colspan attribute and rewrite them with a shortcode
+ **/
 turndownService.addRule("colspan", {
   filter:      node => node.nodeName === "TD" && node.getAttribute("colspan"),
   replacement: (content, node, options) => {
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -7,7 +7,8 @@ const {
   AWS_REGEX,
   BASEURL_SHORTCODE,
   SUPPORTED_IFRAME_EMBEDS,
-  YOUTUBE_SHORTCODE_PLACEHOLDER_CLASS
+  YOUTUBE_SHORTCODE_PLACEHOLDER_CLASS,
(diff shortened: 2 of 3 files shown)


src/constants.js
```diff
@@ -11,8 +11,10 @@ module.exports = {
   AWS_REGEX: new RegExp(
     /https?:\/\/open-learning-course-data(.*)\.s3\.amazonaws.com/g
   ),
-  INPUT_COURSE_DATE_FORMAT: "YYYY/M/D H:m:s.SSS",
-  SUPPORTED_IFRAME_EMBEDS:  {
+  // eslint-disable-next-line no-irregular-whitespace
+  IRREGULAR_WHITESPACE_REGEX: new RegExp(/(\|   )+\|/g),
```
noisecapella, line 15: Is this meant to match any of the whitespace characters between two `|` characters, or do the 3 whitespace characters appear consecutively?

gumaerc: They appear consecutively, although I'm just now realizing that the `+` probably doesn't really do much here.  The middle space is character code 160, aka a non breaking space, and the other two are regular spaces aka character code 32.  I think the issue has to do with table cells with only an `&nbsp;` in them being fed into turndown and the result is the ascii character code for `&nbsp;`, 160.

noisecapella (approved):

src/constants.js
gumaerc: So I was looking at this again, and if you're interested my rationale here was to capture any amount of `|   |`.  Here's a regexr link that demonstrates it: https://regexr.com/5oq0b

Issue #222, opened 2021-03-10
markdown inside of anchor tag isn't rendering


pdpinch:
### Steps to Reproduce

Compare OCW Next: https://ocwnext.odl.mit.edu/courses/8-04-quantum-physics-i-spring-2016/sections/video-lectures/

to OCW Legacy: https://ocw.mit.edu/courses/physics/8-04-quantum-physics-i-spring-2016/video-lectures/

### Expected Behavior

- Link text should be rendered with `<strong>` tags. 

On the legacy site:

```<p><a href="/courses/physics/8-04-quantum-physics-i-spring-2016/video-lectures/part-1"><strong>Part 1: Basic Concepts</strong></a></p>```

### Actual Behavior

- markdown asterisks aren't rendered:

```<p><a href="https://ocwnext.odl.mit.edu/courses/8-04-quantum-physics-i-spring-2016/sections/video-lectures/part-1">**Part 1: Basic Concepts**</a></p>```

### Related Issues

It strikes me this might be the same issue as #221 where a markdown image tag inside a link isn't being rendered.

Issue #221, opened 2021-03-10
markdown rendering error for PDF accessibility statement


pdpinch:
### Steps to Reproduce

Compare OcwNext: https://ocwnext.odl.mit.edu/courses/6-00sc-introduction-to-computer-science-and-programming-spring-2011/sections/unit-3/lecture-21-using-graphs-to-model-problems-part-1/  (3rd bullet under resorces)

to OCW legacy:

https://ocw.mit.edu/courses/electrical-engineering-and-computer-science/6-00sc-introduction-to-computer-science-and-programming-spring-2011/unit-3/lecture-21-using-graphs-to-model-problems-part-1/

### Actual Behavior

The 3rd bullet under resources renders in HTML as:

```<li><a href="https://ocwnext.odl.mit.edu/courses/6-00sc-introduction-to-computer-science-and-programming-spring-2011/sections/unit-3/lecture-21-using-graphs-to-model-problems-part-1/mit6_00scs11_lec21_slides">Lecture slides (![This resource may not render correctly in a screen reader.](/images/inacessible.gif)PDF)</a></li>```

### Expected Behavior

- for most browsers, there is an invisible gif preceding the PDF link
- for screenreaders, the alt text is read "This resource may not render correctly in a screen reader."

The HTML on the legacy site renders as:

```<li><a href="/courses/electrical-engineering-and-computer-science/6-00sc-introduction-to-computer-science-and-programming-spring-2011/unit-3/lecture-21-using-graphs-to-model-problems-part-1/MIT6_00SCS11_lec21_slides.pdf">Lecture slides <span class="nobr">(<img src="/images/inacessible.gif" alt="This resource may not render correctly in a screen reader.">PDF)</span></a></li>```

PR #230, opened 2021-03-18
remove unnecessary escape calls

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/222
Closes https://github.com/mitodl/ocw-to-hugo/issues/221

#### What's this PR do?
This PR removes some code in the `baseurlshortcode` turndown rule that was taking markdown content and running it through `turndownService.escape` before returning.  It was causing markdown formatting characters in links to be pre-escaped before being written into the markdown file.  I believe this was originally done as part of [this](https://github.com/mitodl/ocw-to-hugo/pull/59) PR, but it doesn't seem to be necessary as I tested the course that was having the issue tied to it and the link in question still renders properly.

This change affects a large amount of content, because it's part of the turndown rule that processes all links containing the `baseurl` shortcode.  The best way to verify it is to s3 sync the entire `open-learning-course-data-production` bucket to your local machine, then run the following replacing my path to `open-learning -course-data` with your own:
```
mkdir private/test
node . -i ~/Code/open-learning-course-data -o private/test --rm
```
When you search the output folder at `private/test` for `[\*\*` there should only be results in 2 files, `private/test/6-441-information-theory-spring-2010/content/sections/readings.md` and `private/test/sts-464-cultural-history-of-technology-spring-2005/content/sections/readings.md`.  In both of these cases, the escaped double asterisk seems intentional.

I would also look at the 2 issues that this PR closes and inspect the output of the courses mentioned in those specifically, ensuring that markdown inside of link text is not escaped.  I would also look at the course mentioned in [this](https://github.com/mitodl/hugo-course-publisher/issues/166) issue and make sure there isn't a regression.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "aws-sdk": "^2.671.0",
     "cli-progress": "^3.4.0",
     "dotenv": "^8.2.0",
+    "gfm-escape": "^0.2.0",
     "js-yaml": "^3.13.1",
     "markdown-doc-builder": "^1.3.0",
     "moment": "^2.29.1",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "aws-sdk": "^2.671.0",
     "cli-progress": "^3.4.0",
     "dotenv": "^8.2.0",
-    "gfm-escape": "^0.2.0",
     "js-yaml": "^3.13.1",
     "markdown-doc-builder": "^1.3.0",
     "moment": "^2.29.1",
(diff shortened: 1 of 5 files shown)


noisecapella:
I'm seeing an issue with HTML not being converted to markdown correctly but I don't think it's a breaking change from this PR so I'll file a separate issue for it

noisecapella (approved):

Issue #265, opened 2021-05-03
exit code 1 even if verbose is turned off


gumaerc:
### Steps to Reproduce

Run the full conversion against an entire `open-learning-course-data` bucket with something like `node . -i ~/Code/open-learning-course-data -o private/test --rm`

### Expected Behavior

Even if there are errors during the conversion, the process should exit with code 0 and log the errors in `ocw-to-hugo.error.log`

### Actual Behavior

The process exits with code 1, which stops the entire build on the build server

gumaerc:
this ended up being caused by some confusion I had due to using an outdated version of `ocw-to-hugo`

PR #267, opened 2021-05-03
don't exit with code 1 unless being run in verbose mode

gumaerc:
#### Pre-Flight checklist

- [x] Testing

Closes https://github.com/mitodl/ocw-to-hugo/issues/265

#### What's this PR do?
The HTML to markdown conversion against the entire OCW library of courses is a naturally error prone process that creates a lot of errors.  These errors are logged to `ocw-to-hugo.error.log` using Winston.  The `--verbose` flag will print these errors to the console, and that is the only situation in which `ocw-to-hugo` should exit with a code 1.  This PR removes a call to `process.exit(1)` in the catch block of the main `run` function, which causes the process to exit with code 1 regardless of `--verbose` being turned on.

 - Sync a parsed OCW JSON bucket locally to your computer, such as `open-learning-course-data-rc`
 - Run `ocw-to-hugo` against the entire folder with a command like `node . -i /path/to/open-learning-course-data -o private/output --rm`
 - Run `EXIT_CODE=$?`, then `echo $EXIT_CODE` and make sure that it returns `0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ If the `-c` option has been specified, processing will be filtered by the course
 | `--strips3`       | No  | `true or false` | Strip the s3 base URL from all OCW resources |
 | `--staticPrefix`       | No  | `/path/to/static/assets` | When `--strips3` is set to true, replace the s3 base URL with this string |
 | `--rm` | No | `true or false` | Clear the contents of the path passed with `-o` before the conversion run |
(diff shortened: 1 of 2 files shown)


gumaerc:
Closing this PR as it was a result of some confusion on my part, `ocw-hugo-themes` had an outdated version of `ocw-to-hugo`.

Issue #256, opened 2021-04-09
CHP: Display old course numbers


pdpinch:
Over time, some courses in OCW have been "renumbered." To help with findability, it's important that the old course numbers are maintained. In OCW legacy, these are rendered in HTML as "[new course number] (formerly [old course number]". See, for example, [16.89j S07](https://ocw.mit.edu/courses/aeronautics-and-astronautics/16-89j-space-systems-engineering-spring-2007/).

The data necessary for this is missing from the parsed JSON, until https://github.com/mitodl/ocw-data-parser/issues/137 is fixed. 

#### Acceptance Criteria:

- [ ] Include the former course number parenthetically after the current course number

#### Related issues 

- https://github.com/mitodl/open-discussions/issues/3381 The old course number should be searchable, just like the new one is

pdpinch:
Note that the course 18-443-statistics-for-applications-spring-2015 is an interesting case. The course number in the URL is the _old_ course number. When this is implemented, this course should display the new course number, 18.650, as the primary number. See [OCW legacy](https://ocw.mit.edu/courses/mathematics/18-443-statistics-for-applications-spring-2015/) for reference)

gumaerc:
Note that while https://github.com/mitodl/ocw-to-hugo/pull/279 will close this issue, there is a caveat with the way this is handled right now.  The structure of course number updates in the source data is an array of objects that look like:

```
    "new_course_numbers": [
        {
            "course_type_col": "Undergraduate",
            "new_course_number_col": "18.650",
            "old_course_number_col": "18.443",
            "sort_as_col": "18.650"
        }
    ],
```

Notice that this is an array, and seems to indicate it would support a course being re-numbered multiple times.  However, there are no OCW courses that currently put more than one item in here.  The code added in the above PR handles only a single re-number.  If multiple re-numbers from legacy data are desired in the future, support will have to be added to "crawl" the items in `new_course_numbers` and build a chain of updates.

PR #279, opened 2021-05-12
display updated course numbers

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/256

#### What's this PR do?
In https://github.com/mitodl/ocw-data-parser/pull/140 the `new_course_numbers` field was added, which is a list of updates to either the primary course number (based on `department_number` and `master_course_number`) or any of the linked courses in `extra_course_number`.  This PR adds logic to the `getCourseNumbers` function in `helpers.js` that searches the `new_course_numbers` array for an update to a course number before adding it to the array that it returns.  If an update exists, a string is returned with the new course number followed by "formerly" and the old course number.

Convert a course with a course number update in the parsed json, such as `16-89j-space-systems-engineering-spring-2007` or `18-443-statistics-for-applications-spring-2015` and ensure that there are updated course numbers in the `data/course.json` file in the `course_numbers` array.

While https://github.com/mitodl/ocw-data-parser/pull/140 has been merged and released, https://github.com/mitodl/open-discussions/pull/3398 has not (as of the time of writing) so the parsed JSON on S3 will not be updated.  If this is to be reviewed and tested prior to a merge of the above `open-discussions` PR and a full run of `backpopulate_ocw_data`, the reviewer will need to clone `ocw-data-parser` and parse JSON for these courses manually.


src/data_template_generators_test.js
```diff
@@ -126,9 +133,12 @@ describe("generateDataTemplate", () => {
   })
 
   it("sets the course_number property on the course data template to data parsed from sort_as and extra_course_number properties in the course json data", () => {
-    const expectedValue = helpers.getCourseNumbers(singleCourseJsonData)[0]
-    const foundValue = courseDataTemplate["course_numbers"][0]
-    assert.equal(expectedValue, foundValue)
+    helpers
+      .getCourseNumbers(singleCourseJsonData)
+      .forEach((expectedValue, index) => {
```
noisecapella, line 138: This wouldn't work if there was an error in `getCourseNumbers` where it returned an empty list. Can we calculate the found values and expected values and assert that they are equal outside the loop?

noisecapella (changes requested):

gumaerc:
@noisecapella all set

noisecapella (approved):

PR #290, opened 2021-05-25
Resolve legacy subtitle URLs

### LINKED ISSUE #287: links to legacy site in embedded_media
### Steps to Reproduce

Convert a course that has a video page, like `8-01sc-classical-mechanics-fall-2016`

### Expected Behavior

The links to the various resources in `embedded_media` point to either an external location or an s3 path to an `open-learning-course-data` S3 bucket

### Actual Behavior

Some links point to external services, but some point directly to `ocw.mit.edu` and need to be resolved to an S3 link using `resolveRelativeLinks`

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #287 

#### What's this PR do?
 - Handles URLs starting with `https://ocw.mit.edu` or `http://ocw.mit.edu` are now treated as if they were relative URLs. Previously we just cut off the prefix without making other changes which meant the URLs were likely broken. The link will be resolved to a link to one of our pages, or if it doesn't match anything, the link will be preserved to support `/ans7870` links, `/help` links or other things like that.
 - In addition this also looks through the `technical_location` property of `embedded_media` and resolves URLs there so that the data in the front matter is updated.
 - Looking over the links, I noticed that PDF links were not being generated if the parent_uid was not a course home page. I changed the code to generate links as long as there was a valid parent_uid found, not just for the course home page

Run ocw-to-hugo on master and this branch and compare the directories. 
 - You should see some relative URLs which are updated to be corrected (for example, `/courses/<topic>/<course-id>/` should just be `/courses/<topic>/`.
 - Some new PDF pages will be created and there should be some updated links to them. 
 - Links to `.srt` files should become S3 links. You should be able to copy and paste the link into the browser and download a valid file.

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -14,9 +14,6 @@ const fixLinks = (htmlStr, page, courseData, pathLookup) => {
       ...helpers.resolveYouTubeEmbedMatches(htmlStr, courseData, pathLookup)
     ]
     htmlStr = helpers.applyReplacements(matchAndReplacements, htmlStr)
-
-    // this will be merged into resolveRelativeLinkMatches in a future PR
-    htmlStr = htmlStr.replace(/http:\/\/ocw.mit.edu/g, "")
   }
   return htmlStr
 }
@@ -59,13 +56,43 @@ const generateMarkdownRecursive = (page, courseData, pathLookup) => {
   const coursePageEmbeddedMedia = Object.values(
     courseData["course_embedded_media"]
   )
-    .map(embeddedMedia => {
-      embeddedMedia["course_id"] = courseData["short_url"]
-      embeddedMedia["type"] = "course"
-      embeddedMedia["layout"] = "video"
-      return embeddedMedia
+    .filter(
+      courseEmbeddedMedia => courseEmbeddedMedia["parent_uid"] === page["uid"]
+    )
+    .map(courseEmbeddedMedia => {
+      const embeddedMediaItems = courseEmbeddedMedia["embedded_media"].map(
+        embeddedMedia => {
+          let technicalLocation = embeddedMedia["technical_location"]
+          if (technicalLocation) {
+            const replacement = helpers.resolveRelativeLink(
+              technicalLocation,
+              courseData,
+              pathLookup
+            )
+            if (replacement) {
+              technicalLocation = replacement
+            }
+          }
+
+          if (technicalLocation) {
+            return {
+              ...embeddedMedia,
+              technical_location: technicalLocation
+            }
+          } else {
+            return embeddedMedia
+          }
+        }
+      )
+
+      return {
+        ...courseEmbeddedMedia,
+        course_id:      courseData["short_url"],
+        type:           "course",
+        layout:         "video",
+        embedded_media: embeddedMediaItems
+      }
     })
-    .filter(embeddedMedia => embeddedMedia["parent_uid"] === page["uid"])
   const parents = courseData["course_pages"].filter(
     coursePage => coursePage["uid"] === page["parent_uid"]
   )
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -35,7 +35,7 @@ const generateMarkdownFromJson = (courseData, pathLookup) => {
     {
       name:  "_index.md",
       data:  generateCourseHomeMarkdown(courseData, pathLookup),
-      files: generateCourseHomePdfMarkdown(courseData, pathLookup)
+      files: generatePagePdfMarkdown(courseData, pathLookup)
     },
     ...rootSections.map(
       page => generateMarkdownRecursive(page, courseData, pathLookup),
@@ -248,16 +248,16 @@ const generateCourseHomeMarkdown = (courseData, pathLookup) => {
   }
 }
 
-const generateCourseHomePdfMarkdown = (courseData, pathLookup) => {
+const generatePagePdfMarkdown = (courseData, pathLookup) => {
   /**
-   * Generate markdown files representing PDF viewer pages for PDF's mentioned on the course home page
+   * Generate markdown files representing PDF viewer pages for PDF's mentioned on a page in the site
    */
 
   return courseData["course_files"]
     .filter(
       file =>
         file["file_type"] === "application/pdf" &&
-        file["parent_uid"] === courseData["uid"]
+        pathLookup.byUid[file["parent_uid"]]
     )
     .map(file => {
       const { path: parentPath } = pathLookup.byUid[file["parent_uid"]]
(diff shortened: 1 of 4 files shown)


src/markdown_generators_test.js
```diff
@@ -267,6 +278,21 @@ describe("markdown generators", () => {
...
+        subtitlesCourseJsonData,
+        pathLookup
+      )
+
+      const embeddedMedia = yaml.safeLoad(
+        markdownData[3].media[0].data.split("---\n")[1]
+      )
+      assert.equal(
+        embeddedMedia.embedded_media[6].technical_location,
+        "BASEURL_SHORTCODE/sections/instructor-insights/instructor-interview-meet-the-educator/m_gqolc3clm"
```
gumaerc, line 293: The `baseurl` shortcode is only for use directly within markdown content as `{{< baseurl >}}`.  Since these links are going in front matter, the shortcode won't be applied.  Instead, templates that use this data need to prepend where they are writing the link with `.Site.BaseURL` if it's a relative link.

gumaerc (changes requested):
S3 URL rewriting looks good, just one issue with the relative links.

src/markdown_generators_test.js
noisecapella: Hmm, so that means the link should start with `/sections/...`? How would the template know whether to prepend `.Site.BaseURL` or leave it as is, for example with `/ans7870` urls?

noisecapella: What do you think about having ocw-to-hugo ask for a baseurl and prepend it there instead of in hugo?

gumaerc: I'm not sure how I feel about that, as the content loses its "portability" at that point.  On the production server (right now, at least) we're building fresh content using `ocw-to-hugo` every time.  If we were asking for a `baseUrl` argument in `ocw-to-hugo` and writing that into some links, that markdown wouldn't be usable in build contexts other than the specific context you passed in that baseUrl for.  It's maybe not such a big deal. but it strikes me as going against the grain a bit in the way static site publishing is supposed to work. Either way, I think it's maybe a topic for another PR.

gumaerc: Another option might be rendering the links with `{{< baseurl >}}` in the front, and then when you iterate them in the template you could do something like:

```
{{ range .Params.embedded_media }}
  <a href="{{ .technical_location | markdownify }}">{{ .id }}</a>
{{ end }}
```

This would put the link text through the markdown preprocessor and give you the base URL.

gumaerc: @noisecapella looks like some of those elements have the url stored on `media_location` so you'd have to check that as well, but I think putting the literal shortcode into the url and piping into `markdownify` in the Hugo template should make that work.  Want to give it a shot?

gumaerc: This is also worth looking into: https://gohugo.io/functions/renderstring/

gumaerc: Disregard what I said above about `markdownify`, I just tried it and it doesn't work in this context.  It seems that shortcodes are really only processed on the `.Content` property of markdown files.

gumaerc: So I was digging more into this particular example, and it seems that `m_gqolc3clm` is the `media_location` property on an `embedded_media` element of type `Video-YouTube-Stream`.  In this case, I don't think the URL needs to be rewritten, as it's not really a URL at all.  This is just used to fetch a video from YouTube by its ID.  I don't think we need to be rewriting urls that don't start with `https://ocw.mit.edu`.

gumaerc: The assertion here is actually looking at the URL for `acf8cbbcf3ada9e7b0d390c8b4f8b1e6`, which is originally `https://ocw.mit.edu/courses/global-languages/21g-107-chinese-i-streamlined-fall-2014/instructor-insights/instructor-interview-meet-the-educator/M_gQolc3clM.pdf`.  Shouldn't this be re-written to an S3 URL pointing to the PDF?

gumaerc: The S3 link to this PDF is available in the course's parsed JSON:

```
        {
            "order_index": null,
            "uid": "acf8cbbcf3ada9e7b0d390c8b4f8b1e6",
            "id": "M_gQolc3clM.pdf",
            "parent_uid": "58a6c062c3801594c1010c9bacadde98",
            "title": "3play pdf file",
            "caption": null,
            "file_type": "application/pdf",
            "alt_text": null,
            "credit": null,
            "platform_requirements": "",
            "description": "",
            "type": "OCWFile",
            "file_location": "https://open-learning-course-data-production.s3.amazonaws.com/21g-107-chinese-i-streamlined-fall-2014/acf8cbbcf3ada9e7b0d390c8b4f8b1e6_M_gQolc3clM.pdf"
        },
```

noisecapella: @gumaerc This should be fixed. I changed it to always use a S3 link for PDFs. Regarding `media_location`, I did a grep and couldn't find any URLs for `media_location` which start with `http://ocw.mit.edu` or `https://ocw.mit.edu`

src/markdown_generators.js
```diff
@@ -59,13 +56,44 @@ const generateMarkdownRecursive = (page, courseData, pathLookup) => {
...
+            )
+            if (replacement) {
+              technicalLocation = replacement
+            }
+          }
+
+          if (technicalLocation) {
+            return {
+              ...embeddedMedia,
+              technical_location: technicalLocation
```
gumaerc, line 81: Can we run this through `stripS3`?  On prod `ocw-to-hugo` is run with this and the `staticPrefix` argument to redirect access to S3 resources through Fastly.

gumaerc (changes requested):
Just one more thing:

src/markdown_generators.js
gumaerc: @noisecapella might also want to add a test that spoofs running with `--stripS3` and `--staticPrefix=/coursemedia`

noisecapella: I added a test, what do you think? We already run `stripS3` on `file_location` in `resolveRelativeLinks` and it doesn't look like there is a `technical_location` which has an AWS prefix. Should I still add `stripS3` here too?

gumaerc: At the point `stripS3` is run in `resolveRelativeLinks`, isn't it still prefixed with `ocw.mit.edu`?  At this point, we've transformed it into an S3 link, and should run `stripS3` on it so that if the user has passed in these options, the link is modified as expected.  I'll take another look tomorrow to verify all this.

noisecapella: `stripS3` is run on `file_location` in `resolveRelativeLinks`, and it doesn't look like there are any `file_location`s which start with a `ocw.mit.edu` prefix

gumaerc: I see what you mean now, I was accidentally running with the wrong argument.

gumaerc (approved):
👍 LGTM

Issue #299, opened 2021-06-04
generate yaml config instead of toml


gumaerc:
`ocw-to-hugo` currently generates `menus.toml` files for sites that include links to external sites in their nav menus.  Seeing as we are standardizing on yaml format configuration in `ocw-studio`, we should also use yaml for our Hugo config for consistency.

#### Acceptance Criteria:

- [x] All configuration files generated by `ocw-to-hugo` in the `config` directory should be in the yaml format

PR #300, opened 2021-06-04
move menus.toml to menus.yaml

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/299

#### What's this PR do?
`ocw-studio` is set up to handle yaml configuration files.  For consistency, we should standardize on using yaml configuration across all our Hugo projects.  This PR moves the external menu link config file to a yaml file.

 - Read the readme if you haven't used this utility before
 - Generate markdown for a course with an external link in the nav such as `7-00-covid-19-sars-cov-2-and-the-pandemic-fall-2020`
 - Verify that a valid `menus.yaml` file is generated under `config/_default`

### DIFF
--- a/src/config_generators.js
+++ b/src/config_generators.js
@@ -1,13 +1,29 @@
+const yaml = require("js-yaml")
+
 const { getExternalLinks } = require("./helpers")
 
+const generateHugoConfig = () => {
+  return yaml.safeDump({
+    baseUrl:      "/",
+    languageCode: "en-us",
+    title:        "MIT OpenCourseWare",
+    theme:        ["base-theme", "course"]
+  })
+}
+
 const generateExternalLinksMenu = courseData => {
-  return `${getExternalLinks(courseData).map((externalLink, index) => {
-    return `[[leftnav]]\n\tname = "${externalLink["title"]}"\n\turl = "${
-      externalLink["url"]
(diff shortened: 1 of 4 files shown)


src/config_generators.js
```diff
@@ -1,11 +1,17 @@
...
-    }"\n\tweight = ${index * 10 + 1000}`
-  })}`
+  return yaml.safeDump({
+    leftnav: getExternalLinks(courseData).map((externalLink, index) => {
+      return {
+        name:   externalLink["title"],
+        url:    externalLink["url"],
+        weight: index * 10 + 1000
+      }
+    })
```
aliceriot, line 13: just a style thing, but if you want you can use the single-line style arrow-functions here, both something like




```js
const generateExternalLinksMenu = courseData =>
  yaml.safeDump({
    leftnav: getExternalLinks(courseData).map((externalLink, index) => ({
        name:   externalLink["title"],
        url:    externalLink["url"],
        weight: index * 10 + 1000
    }))
  })
```

aliceriot (approved):
just had one small comment, otherwise looks good

src/config_generators.js
gumaerc: Good call, I always start out with the curly braces in case I need more than one line and then forget to come back and refactor

PR #301, opened 2021-06-07
Handle archived courses

### LINKED ISSUE #274: CHP: "Archived OCW Versions"
As a user looking at a course home page, I'd like to see a list of links to archived versions of the course, available through MIT DSpace. 

In mitodl/ocw-data-parser#74 we added a field other_version_parent_subjects to the parsed JSON. The value of this field can be used as a key to match multiple versions of the current course.

### Design

See [InVision](https://projects.invisionapp.com/share/QFZ8KA9SH2P#/screens/435953154_OCW_Course_Home_Page_Color_10-27-2020_Course_Info_Collapsed_V2), specifically the section with the header "Archived OCW Versions"

Legacy Example: https://ocw.mit.edu/courses/materials-science-and-engineering/3-012-fundamentals-of-materials-science-fall-2005/

### Pre-Requisites

- https://github.com/mitodl/ocw-data-parser/issues/141 parse archived versions from raw JSON data
- add the necessary data into the course data template

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #274 

#### What's this PR do?
Adds markdown about archived courses to the course json data. Archived courses are found by iterating through the course uids in `is_update_of` and parsing the link from `dspace_handle` if it exists. This is done recursively because courses [like this one](https://ocw.mit.edu/courses/political-science/17-40-american-foreign-policy-past-present-and-future-fall-2017/) for example list archived courses which are anywhere in the parent tree. 

This also refactors the pathLookup code a bit so that `byMasterSubject` (renamed to `coursesByMasterSubject`) uses course info in `byUid` instead of storing it in `byMasterSubject`. 

There are 5 commits which are broken up so the reviewer can differentiate between the refactoring steps and the new archived courses feature. All tests should be passing for each commit so it should show that the refactoring didn't change output at any point (hopefully).

First, convert all raw courses to parsed courses using the latest version of ocw-data-parser (`release` is fine). There are some fields which are necessary for ocw-to-hugo to have information needed to construct archive links.

Convert all courses and compare the directories for the output. You should see a new field `archived_courses` in `1-011-project-evaluation-spring-2011` with a list like this:

    [1.011 PROJECT EVALUATION](https://dspace.mit.edu/handle/1721.1/75001) |  SPRING 2005
    [1.011 PROJECT EVALUATION](https://dspace.mit.edu/handle/1721.1/36392) |  SPRING 2003

If you view the [course on the legacy site](https://ocw.mit.edu/courses/civil-and-environmental-engineering/1-011-project-evaluation-spring-2011/) you should see the same two links to the same two courses. They might be in a different order.

Note that this PR increases the running time from 3m:0s to 3m:20s minutes since the refactoring means it takes multiple passes to build the paths. I think that's fine personally since we are trying to minimize memory usage, not running time necessarily, but let me know if you have any thoughts there


src/file_operations.js
```diff
@@ -28,7 +28,15 @@ const makeUidInfo = courseData => {
   // or other places which may need a little bit of context on external items
 
   const types = {
```
gumaerc, line 30: This is a bit of a nitpick, but the name of this function and the name of the constant `types` is a bit confusing.  The function returns a lookup, so why not call the function `makeUidInfoLookup` or something along those lines?  As far as `types`, it looks like originally these objects only contained a `type` property, but now they contain a lot more than that.  Something a bit more descriptive like `uidInfoObjects` might make more sense.

gumaerc (changes requested):
Overall this looks great and functions well.  I ran all courses through and compared the output to master, confirming that the only thing that had changed is that the `achived_versions` array was added.  I just have one nitpick about variable naming:

src/file_operations.js
noisecapella: Sounds good, just committed the change

gumaerc (approved):
:+1: LGTM

PR #314, opened 2021-07-15
Fix issue with a pipe being treated as a string

annagav:
#### Pre-Flight checklist

- [ ] Testing
  - [x] Code is tested

Fix issue with a pipe being treated as a string

#### What's this PR do?
Pipes are transformed into a url encoded string because markdown tables are built with pipes. This PR changes a pipe to a semicolon.
Also wrapped the parameters into quotes, since the new parameter is a full URL and contains slashes and a colon which breaks it on the Hugo side.

Run `node . -i private/input -o private/output --courses private/courses.json --download --rm` , there should be no errors.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -251,7 +251,7 @@ const getYoutubeEmbedCode = media => {
   return youTubeMedia
     .map(
       embeddedMedia =>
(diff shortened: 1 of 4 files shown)


gumaerc (approved):
:+1:

PR #311, opened 2021-07-14
move nav items to config file

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-to-hugo/issues/309

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/364, we added support for generating navigation menus in `ocw-studio`.  This PR adjusts the way `ocw-to-hugo` generates menu items to be more in line with the way we expect `ocw-studio` to generate them.  Menu items have been removed from markdown content front matter and have been moved into a single `menus.yaml` file for all menu entries, including internal and external links.

#### How should this be manually tested?
 - Read the readme if you've never used `ocw-to-hugo` before
 - Generate markdown for any course (you can find some examples in `example_courses.json` on the `master` branch
 - Generate markdown for the same course on the `cg/nav-to-config-file` branch and output to a different folder
 - Compare the menu items in the `master` branch output's markdown files to the ones in the branch output's `menus.yaml` file and ensure that all the same menu items are being generated

PR #320, opened 2021-07-29
add parent_uid to the front matter on sections that have a parent

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to https://github.com/mitodl/ocw-studio/issues/430

#### What's this PR do?
In https://github.com/mitodl/ocw-to-hugo/pull/311, we moved nav menu configuration into a single `menus.yaml` file.  `ocw-studio`'s import functionality that consumes the output of `ocw-to-hugo` is currently configured to use the old front matter based menu entries to determine the parent / child relationships of imported pages.  This PR adds the `parent_uid` property to sections that have a parent so `ocw-studio` can determine this relationship without depending on the menu entries.

 - Read the readme if you have never used `ocw-to-hugo` before
 - Set up `ocw-to-hugo` to download content from S3
 - Convert a course that has a section with children, like `12-001-introduction-to-geology-fall-2013`
 - Inspect the output and make sure pages in the Instructor Insights section have `parent_uid` properties and the UID matches the UID of the Instructor Insights page itself

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -307,6 +307,7 @@ const generatePdfMarkdown = (file, courseData) => {
     type:          "course",
     layout:        "pdf",
     uid:           file["uid"],
+    parent_uid:    file["parent_uid"],
     file_type:     file["file_type"],
     file_location: helpers.stripS3(file["file_location"]),
     course_id:     courseData["short_url"]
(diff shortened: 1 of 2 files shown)


annagav (approved):
works! 👍

PR #336, opened 2021-09-07
Fix broken test for markdown files

noisecapella:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

None

#### What's this PR do?
Fixes a broken test. Most of the test wasn't actually getting run because the test was operating on `ec-711-d-lab-energy-spring-2011` but it was looking at `2-00aj-exploring-sea-space-earth-fundamentals-of-engineering-design-spring-2009`. Because of how the test was written, there were no assertions and therefore the test always passed. I changed it to be a simple assertion of output markdown filenames

N/A

### DIFF
--- a/src/markdown_generators_test.js
+++ b/src/markdown_generators_test.js
@@ -157,101 +157,22 @@ describe("markdown generators", () => {
       }
     }
 
-    it("contains the course home page and other expected sections", () => {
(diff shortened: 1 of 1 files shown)


abeglova (approved):
👍

PR #337, opened 2021-09-07
Create page for each resource

### LINKED ISSUE #335: Export all files and youtube videos as resources
Instead of treating PDF files specially, we should output resource markdown files instead of as a page. All files should output the same markdown with `layout: resource`
 - The metadata for video resources needs to be updated to match this: https://github.com/mitodl/ocw-hugo-projects/blob/349930043fc3045522a9fa36da47c973fe4f1ce6/ocw-course/ocw-studio.yaml#L107
 - File metadata should include `resourcetype` and `file_type` and the other fields already defined in the schema
 - Links to resource pages should be output as shortcodes where possible (ie, everywhere except in course.json). We need to change the link output from `[title text](url)` to `{{< resource uuid-goes-here "title text" >}}`
 - All resource links should go to `/resources/...` with a unique slug filename to avoid name collisions

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #335 
Related to https://github.com/mitodl/ocw-studio/issues/508

#### What's this PR do?
Creates resource pages for all files and youtube media. PDF pages are now links to `/resources/...`

Diffing this vs master won't work since there are many changes as well as changes in file locations. Instead I would just check a few things in a few courses:
 - Convert `18-06-linear-algebra-spring-2010` using this PR
 - Start ocw-hugo-themes pointing to the outputted data. Use this branch: https://github.com/mitodl/ocw-hugo-themes/pull/172
 - Go to http://localhost:3000/pages/video-lectures/ and click on some video links. You should see something very similar to what's on production: https://ocwnext.odl.mit.edu/courses/18-06-linear-algebra-spring-2010/pages/video-lectures/. Videos should work fine. I've noticed captioning doesn't work but that's a separate issue https://github.com/mitodl/ocw-data-parser/issues/155
 - Go to http://localhost:3000/resources/18-06classroom/. You should see a link where you can download an image file of a classroom. Note that the image is linked on http://localhost:3000/pages/instructor-insights/ but it won't show up due to #352
 - Go to http://localhost:3000/resources/mit18_06s10_pset1_s10_soln/. You should see a PDF render properly. If you go to /pages/assignments and click the first link that says "PDF", it should go to this page.


src/helpers.js
```diff
@@ -291,68 +291,86 @@ const getYoutubeEmbedCode = media => {
...
-      parentItem,
-      itemsLookup,
-      courseUid,
-      pathLookup,
-      uidInfoLookup
-    )
     if (!pathLookup[parentUid]) {
-      loggers.fileLogger.error(
-        `Unable to find path for ${parentUid}, parent of ${uid}`
+      buildPathRecursive(
```
gumaerc, line 340: Can we add a test that hits this?  Codecov is reporting it as unhit.

gumaerc (changes requested):
This looks great, there was only one thing I noticed that I think should be changed as well as some missing test coverage in `buildPathRecursive`:

noisecapella:
@gumaerc I did some refactoring of the tests and I added a course to hit the edge cases. Surprisingly `buildPathsRecursive` was never actually being called recursively, which is a little strange, but I guess things are laid out in the JSON files where the parents always come first so there is no need to do that. It wasn't getting hit even when running it on all courses so I removed it.

gumaerc:
> @gumaerc I did some refactoring of the tests and I added a course to hit the edge cases. Surprisingly `buildPathsRecursive` was never actually being called recursively, which is a little strange, but I guess things are laid out in the JSON files where the parents always come first so there is no need to do that. It wasn't getting hit even when running it on all courses so I removed it.

Interesting.  I suspect this might be because before we started respecting the `list_in_left_nav` property we were generating 3rd tier nav links, but now we're not.  I think the only courses that would have potentially called it recursively would have been edX import courses.

src/markdown_generators.js
```diff
@@ -260,35 +191,98 @@ const generateCourseSectionMarkdown = (page, courseData, pathLookup) => {
...
     title:         file["title"],
     description:   file["description"],
-    layout:        "pdf",
     uid:           file["uid"],
-    parent_uid:    file["parent_uid"],
+    resourcetype:  getResourceType(file["file_type"]),
     file_type:     file["file_type"],
-    file_location: helpers.stripS3(file["file_location"])
+    file_location: helpers.stripS3(file["file_location"]),
+    layout:        "resource"
```
gumaerc, line 234: Still seeing `layout: "resource"` here.  Can we remove it here too?

gumaerc (changes requested):
Just one last spot with `layout: "resource"`

src/markdown_generators.js
noisecapella: Sure, should be gone now

gumaerc:
@noisecapella I just merged https://github.com/mitodl/ocw-to-hugo/pull/366, so after a quick rebase this should be good to go.

gumaerc (approved):
:+1: LGTM

PR #342, opened 2021-09-09
Refactor path handling

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Related to #337 

#### What's this PR do?
This was part of #337 but I broke it out to be an easier review since this is mostly just a refactor with a couple exceptions. This does a few things:
 - Course feature links don't start with a `/` anymore. For example `/pages/projects` is now `pages/projects`. I think this is more intuitive as a relative link, it simplifies the cases we need to consider for creating links, and it should require only a tiny change in the template to handle it
 - The `rootRelativeToDocumentRelative` string replace code is removed. Instead we are making links with or without the slash prefix in `constructInternalLink`
 - `constructInternalLink` is meant to centralize the logic on what links should be rendered when, if shortcodes are allowed to be used or if the link should be relative to the course home page. #337 will expand this and add a resource shortcode which will be included as appropriate
 - PDF links are now created in `pathLookup`, with a new property `unalteredPath` which is used in some cases to resolve relative links
 - The whole path of a resource is matched, and in some cases the pdfs we export have different `technical_location` urls. I tried a few links from different courses where they differed and they still pointed to identical files so I don't know if the end user will see any improvement here. However you might notice that the uid of the course file is included in the `technical_location` now so I think it is probably resolving more correctly than before.

Convert all courses on master, then convert all courses on this PR, and compare the two directories. You probably won't be able to check everything but pick a few files which are different and verify that the changes you see make sense.

You should see some differences:
 - course feature links now start with `pages` instead of `/pages`
 - `technical_location` urls will be different in some cases
 - links to other courses in course descriptions now start with `/courses/...` instead of `courses/...` which should fix those links
 - `/ans7870` links in course descriptions are fixed and now start with a slash

I noticed a couple differences where the old behavior is broken so I don't consider these regressions:
 - A pdf link was updated and doesn't work, but it was also broken before so I am not counting this as a regression https://github.com/mitodl/ocw-to-hugo/issues/343
 - Another link was changed but that link was also broken beforehand #344

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -15,7 +15,10 @@ const {
   FILE_TYPE,
   INPUT_COURSE_DATE_FORMAT,
   YOUTUBE_SHORTCODE_PLACEHOLDER_CLASS,
-  ROOT_RELATIVE_REGEX
+  ROOT_RELATIVE_REGEX,
+  EMBEDDED_MEDIA_PAGE_TYPE,
+  RESOURCE_PLACEHOLDER,
+  COURSE_TYPE
 } = require("./constants")
 const loggers = require("./loggers")
 const runOptions = {}
@@ -188,8 +191,10 @@ const getCourseFeatureObject = (courseFeature, courseData, pathLookup) => {
     featureObject["url"] = resolveUidForLink(
       url,
       courseData,
-      pathLookup
-    ).replace(BASEURL_PLACEHOLDER_REGEX, "")
+      pathLookup,
+      false,
+      true
+    )
   }
   return featureObject
 }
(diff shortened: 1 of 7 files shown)


src/helpers.js
```diff
@@ -15,7 +15,10 @@ const {
   FILE_TYPE,
   INPUT_COURSE_DATE_FORMAT,
   YOUTUBE_SHORTCODE_PLACEHOLDER_CLASS,
-  ROOT_RELATIVE_REGEX
+  ROOT_RELATIVE_REGEX,
```
gumaerc, line 18: This and a couple other constants here aren't used anymore.  Can we remove unused imports?

gumaerc (changes requested):
I spent a lot of time today going through the diff between the output from this and master.  This is unquestionably a better way to handle the questions "is this relative to the course home page?" or "can we use shortcodes here?"  I didn't notice any differences besides the ones you noted, but the transcript `technical_location` changes are interesting.  For instance, this URL:

https://open-learning-course-data-production.s3.amazonaws.com/18-06sc-linear-algebra-fall-2011/da7edc26bc12b2aca08bbeff1dc2f25f_My5w4MXWBew.pdf

and this one:

https://open-learning-course-data-production.s3.amazonaws.com/18-06sc-linear-algebra-fall-2011/e9f8e448838a42ea9a9e5dd9ab6628b8_My5w4MXWBew.pdf

...have different UUID's but are exactly the same file.  In that regard, there might be something going on in `ocw-data-parser` (or even further up the chain) that is adding duplicate files?  Anyway, I see no issues with this refactor overall and the changes to the unit tests are very sensible.  I just noticed one spot where there are some unused imports:

gumaerc (approved):
:+1: LGTM

PR #362, opened 2021-09-16
Add dashes to exported uuids

### LINKED ISSUE #361: Add dashes to all exported uuids
ocw-studio has dashes included for every uuid in the exported output, so we should do the same here. Currently we pass through uid values from Plone which do not have dashes included

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #361 
Rebased on #337 

#### What's this PR do?
Adds dashes to all exported uuids to be consistent with how we store and use them in ocw-studio

Run ocw-to-hugo on all courses, then run `grep -P "[^/][a-f0-9]{32}\b" . -R` or a similar command. You should see only a relatively small number of results, around 100-200 or so, and all uuids should be unrelated, used in other URLs for example. If you run the same command on master you will get ~150k results

### DIFF
--- a/src/config_generators_test.js
+++ b/src/config_generators_test.js
@@ -30,13 +30,13 @@ describe("generateMenuItems", () => {
     const expectedValue = yaml.safeDump({
       leftnav: [
         {
-          identifier: "b6c8c090d7079126837f7dda4af627c7",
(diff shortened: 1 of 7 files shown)


gumaerc (approved):
:+1:

PR #399, opened 2021-10-29
Update level, year for legacy template

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Related to #398 

#### What's this PR do?
I overlooked the legacy course JSON data when updating how we handle levels

#### How should this be manually tested?
Convert a course, and compare the directories. The course legacy JSON file should have a simple array of levels instead of a `{ level, url }` data structure.

PR #402, opened 2021-11-01
Change term to be only the semester without year

noisecapella:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Related to #399 

#### What's this PR do?
Updates term to be only the "Fall", "Summer", "Spring", or "January IAP" part of the string. The year is not included anymore since it is now in its own field.

Convert a course and verify that only the term has changed, not including the year part


abeglova (approved):
LGTM

PR #409, opened 2021-11-17
Add image metadata

noisecapella:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of https://github.com/mitodl/ocw-hugo-themes/issues/260

#### What's this PR do?
Adds `image_metadata` for image resources. For the course image or thumbnail it will use the same alt text that we use in the legacy JSON

Convert `1-00-introduction-to-computers-and-engineering-problem-solving-spring-2012`. Look at the course image and thumbnail uids in the `course.json` file. The uids will match `1-00s12.md` and `1-00s12-th.md`. Look in both of those files. You should see a `image_metadata` section which says "One big and one small circuit board" for the alt text, and "This phidget kit features one of the many tools" in the caption

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -70,7 +70,7 @@ const generateResourceMarkdown = (courseData, pathLookup) => {
 
       return {
         name: `${resourcePath}.md`,
-        data: generateResourceMarkdownForFile(file, courseData)
+        data: generateResourceMarkdownForFile(file, courseData, pathLookup)
       }
     } catch (err) {
       loggers.fileLogger.error(err)
@@ -227,7 +227,7 @@ const getResourceType = mimeType => {
   }
 }
 
-const generateResourceMarkdownForFile = (file, courseData) => {
+const generateResourceMarkdownForFile = (file, courseData, pathLookup) => {
   /**
   Generate the front matter metadata for a PDF file
   */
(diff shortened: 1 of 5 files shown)


src/markdown_generators.js
```diff
@@ -227,19 +227,41 @@ const getResourceType = mimeType => {
...
+    const alt = isCourseImage
+      ? courseData["image_alternate_text"]
+      : file["alt_text"]
+    const caption = isCourseImage
+      ? courseData["image_caption_text"]
+      : file["caption"]
+
+    frontMatter["image_metadata"] = {
+      ["image-alt"]: alt || "",
+      caption:       caption || "",
```
gumaerc, line 260: Can we wrap this in `html2markdown(fixLinks(`?  I tested converting all courses and a number of the captions not only have HTML in them but also have `resolveuid` links in them.  Raw HTML in these fields will come through as raw text if this is rendered in Hugo.

gumaerc (changes requested):
Looks good overall, just one concern about the `caption` field:

src/markdown_generators.js
noisecapella: Sure, I'll also do this for the legacy JSON

gumaerc (approved):
:+1: LGTM

PR #418, opened 2021-11-30
parse multiple choice

abeglova:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/ocw-hugo-themes/issues/215
)
#### What's this PR do?
This pr parses interactive multiple choice questions as a series of shortcodes

Follow the steps in https://github.com/mitodl/ocw-hugo-themes/pull/290

### DIFF
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -421,6 +421,174 @@ turndownService.addRule("div_with_class", {
   }
 })
 
+turndownService.addRule("multiple_choice_question", {
+  filter: (node, options) => {
+    if (node.nodeName === "DIV" && node.getAttribute("class")) {
+      if (
+        node.getAttribute("class") === "problem_question" &&
+        node.getElementsByClassName("problem_radio_input")
+      ) {
+        return true
+      }
+    }
+    return false
+  },
+  replacement: (content, node, options) => {
+    const questionId = node.getAttribute("id")
+    let questionMarkdown = Array.from(
+      node.getElementsByTagName("fieldset")[0].childNodes
+    )
+      .filter(child => {
+        if (
+          child.nodeName === "DIV" &&
+          child.getAttribute("class").includes("choice") &&
+          child.getElementsByClassName("problem_radio_input")
+        ) {
+          return true
+        }
+      })
+      .map(choice => {
+        const isCorrect =
+          choice
+            .getElementsByClassName("problem_radio_input")[0]
+            .getAttribute("correct") === "true"
+        return `{{< quiz_choice isCorrect="${isCorrect}" >}}${turndownService.turndown(
+          choice.outerHTML
+        )}{{< /quiz_choice >}}`
+      })
+      .join("\n")
+    questionMarkdown = `{{< quiz_choices >}}${questionMarkdown}{{< /quiz_choices >}}`
+    const solution = helpers.getNextSibling(node, "div.problem_solution")
+
+    let solutionMarkdown = `{{< quiz_solution / >}}`
+
+    if (solution) {
+      solutionMarkdown = `{{< quiz_solution >}}${turndownService.turndown(
+        solution.outerHTML
+      )}{{< /quiz_solution >}}`
+    }
+
+    questionMarkdown = `${questionMarkdown}\n${solutionMarkdown}`
+
+    return `{{< quiz_multiple_choice questionId="${questionId}" >}}${questionMarkdown}{{< /quiz_multiple_choice >}}`
+  }
+})
+
+turndownService.addRule("multiple_choice_questions_widget", {
+  filter: (node, options) => {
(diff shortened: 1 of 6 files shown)


src/turndown.js
```diff
@@ -421,6 +421,174 @@ turndownService.addRule("div_with_class", {
...
+        ) {
+          return true
+        }
+      })
+      .map(choice => {
+        const isCorrect =
+          choice
+            .getElementsByClassName("problem_radio_input")[0]
+            .getAttribute("correct") === "true"
+        return `{{<choice isCorrect="${isCorrect}">}}${turndownService.turndown(
```
gumaerc, line 455: Minor formatting nitpick:

Here and other places where you are calling shortcodes, can we make sure there are spaces on either side of the decorators every time?  I'm not sure how much this matters, but every place I look at in the Hugo documentation that shows shortcodes has a space after the opening decorator and before the closing.  So this one for example would be :

`{{< choice isCorrect="${isCorrect}" >}}`... etc.

Our code formatter isn't really capable of handling situations like this where we are defining code in strings like this.  It may be unnecessary, but this is the way I've always done it because it's the way I see shortcodes called in all of Hugo's documentation.  With closing tags it would be:

`{{< /choice >}}`

Examples of what I'm talking about can be found here: https://gohugo.io/templates/shortcode-templates/

src/turndown.js
```diff
@@ -421,6 +421,174 @@ turndownService.addRule("div_with_class", {
...
+    if (node.nodeName === "SCRIPT" && node.textContent.includes("quizMulti")) {
+      return true
+    }
+    return false
+  },
+  replacement: (content, node, options) => {
+    const jsParser = require("acorn")
+    const walk = require("acorn-walk")
+    const nodeText = node.textContent.replace(
+      "// There was an extra comma at the end of multiList array.",
```
gumaerc, line 488: I was going to ask about this, then I searched through the parsed JSON and saw how often it appears. :joy:

src/turndown_test.js
```diff
@@ -311,4 +311,105 @@ describe("turndown", () => {
...
+              },
+              {
+                ques: "In the long run, firms should decide to shut down if what condition holds?",
+                ans: "Price is less than both average cost and average variable cost.",
+                ansSel: ["Price is less than average variable cost.", "Price is less than average cost.", "Price is less than average fixed cost."],
+                ansInfo: "In the long run, all costs are variable, and thus average variable cost and average cost are equivalent. The firm will shut down if price is less than average cost. Average fixed cost is not a relevant concept in the long-run, because all costs are considered to be variable."
+              }
+            ]
+          };
+          var options = {
```
gumaerc, line 366: Do we need to do any parsing of these options that seem to appear on each of the quizzes?  Things like `animationType` probably don't matter, but I'm wondering if the `title` property needs to be displayed somewhere or if any of these other options need to be taken into account.  Feel free to make a separate issue for this, as we may not have all the requirements at the moment.  I figured I would mention it though because they do seem to appear on every quiz and are being passed into this "jQuizMe" library.  If they change the behavior we should note that and potentially implement whatever the options are doing.

gumaerc (changes requested):
Overall this works well and I only have some minor comments.  I have a few concerns about the naming of the shortcodes that I'll address in the `ocw-hugo-themes` review.

gumaerc:
Also to note for the future, I was able to figure this out but it's helpful in the testing instructors (either here or in the linked `ocw-hugo-themes` PR) to note if you're adding a new external dependency.  It requires a fresh `yarn install` to be run before testing.  Turndown is bad about swallowing errors and I didn't figure this out until I ran the conversion and saw that I was getting raw HTML on the other end.  I've been thinking about making some changes to the PR /  issue templates so I'll keep this in mind.

src/turndown_test.js
abeglova: The title is not actually displayed anywhere is original ocw. Here is the page where the example was taken from:
https://ocw.mit.edu/courses/economics/14-01sc-principles-of-microeconomics-fall-2011/unit-3-producer-theory/competition-ii/

The rest of the options are settings for the quiz widget. We aren't actually going to display the widget in ocw-next, just show the questions in the page.

abeglova:
Sorry about forgetting to mention that `yarn install` is needed!

src/turndown.js
abeglova: Yeah, it appears a lot, and also breaks the parsing because `textContent` gets rid of whitespace so everything after `//` gets interpreted as a comment. Not sure if there is a less hacky way to fix this, but this should be fine since there will definitely not be any new courses that use the `jQuizMe` library

abeglova:
back to you @gumaerc, thanks for the review

gumaerc (approved):
:+1:

PR #419, opened 2021-11-30
update readme

alicewriteswrongs:

none

#### What's this PR do?

I just cloned this repo for the first time in a while on a new computer, and realized that the README wasn't accurate. In particular, it tells the user to set a `-d` flag to tell `ocw-to-hugo` to download courses, but the flag is actually `--download`. So I updated the README to reflect that, and did a little bit of formatting too.

The proposed changes should make sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,10 @@
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


gumaerc (approved):
:+1: Good catch!

PR #424, opened 2021-12-13
add uid to course.json and course_legacy.json

abeglova:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/ocw-studio/issues/870

#### What's this PR do?
In order to be able to import course uids to ocw-studio, we need to add the uid to course_legacy.json
 
Run
```
node . -i private/input -o private/output -c course_json_examples/example_courses.json --download
```
Go to the /data subfolder for any course in your output.

Verify that `uid` is set in course_legacy.json and course.json


src/data_template_generators.js
```diff
@@ -47,7 +47,8 @@ const generateDataTemplate = (courseData, pathLookup) => ({
   ),
   open_learning_library_versions: helpers.getOpenLearningLibraryVersions(
     courseData["open_learning_library_related"]
-  )
+  ),
+  uid: courseData["uid"]
```
gumaerc, line 51: Can we wrap this with `helpers.addDashesToUid`?  The UUID's that come out of the parsed Plone data are just one string as you see and standard UUID's have dashes in specific spots, this function fixes that.

gumaerc (changes requested):
Just one issue:

gumaerc (approved):
:+1:

PR #438, opened 2022-01-26
make turndown escape angle brackets

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#### What's this PR do?
Some courses contain `&lt;` and `&gt;` entities in their HTML. Turndown does not, by default, escape these. This results in problems for rendering, especially in cases like `&lt;a dog &gt; where the resulting text (`<a dog>`) might be interpretted as a tag.

With this commit, we monkey-patch turndown's escpaing method to escape **opening** angle brackets. (Monkey-patching `TurndownService.prototype.escape` seems like the quasi-endorsed way to do this. See https://github.com/mixmark-io/turndown#escaping-markdown-characters.)

Only escape opening angle brackets (not closing) because Showdown treats escaped closing brackets as literal `\>`.

Using `ocw-to-hugo`, migrate a course containing `&lt;`  in the HTML text and observe that on master they become `"<"` but on this branch they become `\<`.

I used Ancient Philosophy from 2004...
```
# private/courses.json
{
    "courses": [
        "24-200-ancient-philosophy-fall-2004"
    ]
}
# Then
node . -c private/courses.json --download -i private/in -o private/out
```
Then
```
private/out/24-200-ancient-philosophy-fall-2004/content/pages/assignments.md
```
contains some angle brackets.

### DIFF
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -1,4 +1,4 @@
-const TurndownService = require("turndown")
+const OriginalTurndownService = require("turndown")
 const turndownPluginGfm = require("turndown-plugin-gfm")
 const { gfm, tables } = turndownPluginGfm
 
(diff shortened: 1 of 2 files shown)


src/turndown.js
```diff
@@ -15,6 +15,43 @@ const {
...
+  escape(string) {
+    /**
+     * Regex to match
+     *  1) an even number of backslashes,
+     *  2) not preceeded or followed by a backslash
+     *  3) followed by <
+     *
+     * (2) precludes from erroneously matching an odd number of backslashes.
+     */
+    const UNESCAPED_OPENING_ANGLE_BRACKET = /(?<!\\)(?:\\\\)*(?!\\)</g
```
ChristopherChudzicki, line 48: We could potentially just use `/</g`. Originally I going to escape both `<` and `>`, but it turned out Showdown did not like the escaped `>`, so I removed them.

The more complicated regex was mostly to avoid double-escaping `>`. (Turndown DOES escape `>` when they occur at the beginning of a line.) But... not escaping escaped stuff seemed good, so I left it.... 🤷

src/turndown.js
```diff
@@ -15,6 +15,43 @@ const {
...
+   * e.g. "&lt;a dog&gt;"" turns into "<a dog>", which can be erroneously
+   * interpretted as an anchor tag. [Aside: Turndown does escape angle brackets
+   * at the beginning of a line, since those are markdown block quotes.]
+   *
+   * Below, we monkey-patch the escape method to escape OPENING angle-brackets
+   * only, unless they are already escaped. Why only opening?
+   *  - According to the GFM spec (https://github.github.com/gfm/#backslash-escapes)
+   *    escaping opening and closing brackets should be safe
+   *  - But showdown, which we use in OCW Studio, renders "\>" as "\>" instead
+   *    of ">".
```
alicewriteswrongs, line 33: :/ well that's not so nice of showdown!

alicewriteswrongs (approved):
looks good to me!

Issue #464, opened 2022-02-17
Code formatting


pdpinch:
Some courses include code snippets. 

These should be rendered with appropriate HTML tags, and the styling should use a monospaced font. 

example 1:
legacy: https://ocw.mit.edu/courses/electrical-engineering-and-computer-science/6-006-introduction-to-algorithms-fall-2011/syllabus/software/
nextgen: https://ocwnext.odl.mit.edu/courses/6-006-introduction-to-algorithms-fall-2011/pages/syllabus/software/#ide
markdown: https://github.mit.edu/mitocwcontent/6.006-fall-2011/blob/main/content/pages/syllabus/software.md

example 2:
legacy:
nextgen: https://ocwnext.odl.mit.edu/courses/6-006-introduction-to-algorithms-fall-2011/pages/readings/binary-search-trees/

MAbdurrehman12:
> Where are you seeing the inline styling?

Initially, I saw it in the source code (by inspecting element) of the legacy course and after pulling the course from s3, the code snippets seem to have this inline-style.
e.g:   `<span style="font-family: Courier New,Courier;">sudo apt-get install -y python2.7 python-profiler</span>`

> Instead of inline styling it would be better to convert them to the correct markdown

This certainly is better, thank you

gumaerc:
@MAbdurrehman12 This seems like a good strategy.  As you've probably noticed, there are already turndown rules labeled `codeblockfix` and `inlinecodeblockfix`.  These rules catch the typical code block HTML elements (`pre`, `kbd`, `tt`, `samp`) and notate them in markdown as code.  I think if there's some formatting in legacy OCW that's expecteding a span to be formatted with Courier New or Courier, code block notation will accomplish that regardless of the context so this should be fine.  If we find some examples of places where this style is set but the code block formatting doesn't look right, I think we can fix those one-off cases manually.

PR #469, opened 2022-02-22
fix: turndown rules added for code snippets

MAbdurrehman12:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

- Closes https://github.com/mitodl/ocw-to-hugo/issues/464

#### What's this PR do?
- Adds `turndown rule` for code blocks that are wrapped in `<pre>` element.
- Adds `turndown rule` for 1 liner code snippet. 

- Setup `ocw-to-hugo` locally. 
    - Detailed explanation is in the [documentation](https://github.com/mitodl/ocw-to-hugo) but here are few points. 
    - I created a `course.json` file in `private/input` directory so it is ignored by git.
    - Place any course which has code snippets/code blocks in an array named `courses` in `course.json`. (Course example: [6-006-introduction-to-algorithms-fall-2011](https://ocwnext.odl.mit.edu/courses/6-006-introduction-to-algorithms-fall-2011/pages/syllabus/software/#ide))
- Run the command `node . -i private/input -o *output directory*  -c private/input/courses.json`
- Build and run this course in `ocw-hugo-themes`
- Verify that all the code snippets are properly wrapped in their respective elements. i.e: `<pre>`, `<code>`
- Verify that there is no side-effect on any other content.

- Note: Styling of `code blocks` might need to be changed in `ocw-hugo-themes`

image

image

image

### DIFF
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -136,7 +136,7 @@ turndownService.addRule("codeblockfix", {
       }
       return content
     }
-    return `\`\`\`${content}\`\`\`\n`
+    return `\n\n\`\`\`${content}\n\`\`\`\n\n`
   }
 })
 
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -170,6 +170,29 @@ turndownService.addRule("inlinecodeblockfix", {
 })
 
 /**
+ * turn <pre> elements into code blocks
+ */
+turndownService.addRule("code_block", {
+  filter:      node => node.nodeName === "PRE",
+  replacement: (content, node, options) => {
+    return `\`\`\`${content}\`\`\`\n`
+  }
+})
+
+/**
+ * In legacy course, 1 liner code snippets are <span> elements having a particular inline style
+ * For details, see https://github.com/mitodl/ocw-to-hugo/issues/464
+ */
+turndownService.addRule("one_liner_code", {
+  filter: node =>
+    node.nodeName === "SPAN" &&
+    node.getAttribute("style") === "font-family: Courier New,Courier;",
+  replacement: (content, node, options) => {
+    return `\`${content}\`\n`
+  }
+})
+
+/**
  * Build anchor link shortcodes
  **/
 turndownService.addRule("anchorshortcode", {
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -136,7 +136,12 @@ turndownService.addRule("codeblockfix", {
       }
       return content
     }
-    return `\n\n\`\`\`${content}\n\`\`\`\n\n`
+    // if string starts with a 'tab' then it is converted into a code block while MD to HTML 
+    // so no need to make it a code block here
+    if (content.startsWith("\t") || content.startsWith("    ")) {
+      return content
+    }
+    return `\n\n\`\`\`\n${helpers.removeLeadingBackslash(content)}\n\`\`\`\n\n`
   }
 })
 
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -136,7 +136,7 @@ turndownService.addRule("codeblockfix", {
       }
       return content
     }
-    // if string starts with a 'tab' then it is converted into a code block while MD to HTML 
+    // if string starts with a 'tab' then it is converted into a code block while MD to HTML
     // so no need to make it a code block here
     if (content.startsWith("\t") || content.startsWith("    ")) {
       return content
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -124,18 +124,19 @@ turndownService.addRule("tfoot", {
  * fix some strangely formatted code blocks in OCW
  * see https://github.com/mitodl/hugo-course-publisher/issues/154
  * for discussion
+ * also turns <pre> elements into code blocks
  */
+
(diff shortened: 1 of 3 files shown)


src/turndown_test.js
```diff
@@ -180,11 +180,29 @@ _italics wrapped in a div_
...
+      "```import time  \nprint('I am going to sleep :)')  \ntime.sleep(1000)```"
+    )
+  })
+
   it("should not get tripped up on problematic code blocks", async () => {
     const problematicHTML =
       "<pre><span><code>stuff\nin\nthe\nblock</span></pre>"
     const markdown = await html2markdown(problematicHTML)
-    assert.equal(markdown, "```\nstuff\nin\nthe\nblock\n```")
+    assert.equal(markdown, "````stuff\nin\nthe\nblock````")
```
MAbdurrehman12, line 205: This test was failing and I also didn't want to change the function so I had to update it's comparison value here, not sure whether its correct.
@alicewriteswrongs Did you write this? If yes, then is it correct or I did something wrong?

gumaerc: @MAbdurrehman12 This was introduced in [this](https://github.com/mitodl/ocw-to-hugo/pull/58) PR.  I think that if you're seeing this test fail, the original problem this was meant to solve is not being solved anymore.  I think the reason for this is that your turndown rule simply filters on all `pre` elements whereas the other rule applies to `pre` elements that also have a `span` as their first child.  It would seem that your rule that simply filters on `pre` is being selected instead of the other rule.  What I would do here is combine both sets of logic into one turndown rule.  The filter for the turndown rule itself should only be `node.nodeName === "PRE"`, and then in the replacement rule apply the logic from the replacement rule on `codeblockfix` if `node.firstChild && node.firstChild.nodeName === "SPAN"` is true.  This would ensure that only one rule is followed in both scenarios.

src/turndown.js
```diff
@@ -169,6 +169,29 @@ turndownService.addRule("inlinecodeblockfix", {
   }
 })
 
+/**
+ * turn <pre> elements into code blocks
+ */
+turndownService.addRule("code_block", {
+  filter:      node => node.nodeName === "PRE",
```
gumaerc, line 176: We should combine this rule and the `codeblockfix` rule, leaving the filter as you have it here and matching on `pre` elements, then in the replacement rule having a special condition for `node.firstChild && node.firstChild.nodeName === "SPAN"`.  I'm fairly certain an element can only be matched to one turndown rule, and the case `codeblockfix` was meant to be solved is now being caught by this rule and not handled in the intended way.

gumaerc (changes requested):
I made a comment above about this, but just to put it in an official review:

src/turndown_test.js
MAbdurrehman12: Thanks for pointing this out @gumaerc. I'll update the PR

src/turndown.js
MAbdurrehman12: Yeah, this makes sense.
Thanks

pdpinch:
Could I see some before/after examples, so I understand how the legacy HTML is getting converted into markdown?

We should also consult with @alicewriteswrongs about how this might interact with CKEditor. We don't need to do it now, but I'd like to know if we will be able to add a "code" button in the future for authoring new courses.

MAbdurrehman12:
> Could I see some before/after examples, so I understand how the legacy HTML is getting converted into markdown?


1 liner code:
Before:
image

After: 
image

Code block:

Before:
image

After:
image


Styling of all this has been changed in another PR in ocw-hugo-themes

pdpinch:
Can I see the markdown, before and after?

MAbdurrehman12:
> Can I see the markdown, before and after?


So for 1 liner code: 
Before: 
image
After:
image

For code block:
It's the same but with three `backticks` (i.e: ```) instead of one, 
but I just noticed an issue in code-block,  
image
 is being appended in some places inside code. 
 Not sure if this is an actual issue as I guess it's the same as three `backtick`, 
https://wikimass.com/html/entities/grave-accent-like

pdpinch:
Thanks for the code example. I'm concerned about the HTML encoding to `&grave;` but I'm not sure if it's a problem.

MAbdurrehman12:
> I'm concerned about the HTML encoding to `&grave;` but I'm not sure if it's a problem. 


I guess it is a problem as &grave; is not being rendered and it's visible in the output.
Anyhow, I have fixed it but forgot to change my own test which is failing at the moment, I'll also change it shortly and the PR is up for review again

MAbdurrehman12:
@alicewriteswrongs @gumaerc Whenever you have time, can you please review this?

mbertrand (approved):
All the code blocks I found for 6-006-introduction-to-algorithms-fall-2011 look good to me now with the latest commit, will approve the related themes PR as well. 👍

gumaerc (approved):
:+1: LGTM

Changes look good to the code block rule, and good catch with that weird styling on the inline code blocks.  I ran my usual test of converting the entire library of OCW courses and searching through the output.  Every code block I saw looked to be formatted correctly.

Issue #473, opened 2022-02-14
some pages in 7.00sc Fundamentals of Biology don't build correctly


pdpinch:
### Steps to Reproduce

view https://ocw-published.odl.mit.edu/courses/7-01sc-fundamentals-of-biology-fall-2011/pages/biochemistry/macromolecules-lipids-carbohydrates-nucleic-acid/

### Expected Behavior

should have all the content from https://ocw.mit.edu/courses/biology/7-01sc-fundamentals-of-biology-fall-2011/biochemistry/macromolecules-lipids-carbohydrates-nucleic-acid/

### Screenshot or Screencast

![image](https://user-images.githubusercontent.com/430126/153926774-4077f7b4-c127-4c91-b04c-6497b6a4feb9.png)

### Related issues

The content loads in studio, but the page doesn't build. This page contains a "check yourself" javascript, so work done in mitodl/ocw-hugo-themes#215 might be relevant.

gumaerc:
@abeglova I dug into this this morning and found that the issue was happening in the `multiple_choice_questions_widget` turndown rule.  There were a number of times parsing the content of `dataSubstring` as JSON failed.  Here's an example:

```
{ "multiList": [ { "ques": 'Which of the following statements is true for phospholipid molecules?<ol type="a">
  <li>The polar end of phospholipids would contain carbon and phosphorous and oxygen.</li>
  <li>The non-polar end of phospholipids would contain almost exclusively carbon and hydrogen.</li>
  <li>The polar end of phospholipids would form hydrogen bonds with water.</li>
  <li>The non-polar end of phospholipids associate with the cytoplasm of the cell.</li>
</ol>', "ans": "a,b,c", "ansSel": ["a,b,c, and d", "b,c,d", "a,c"], "ansInfo": "" }, { "ques": 'Which of the following
statements is true for carbohydrate molecules? <ol type="a">
  <li>The general structure can abbreviated as (CH<sub>2</sub>O)n.</li>
  <li>A disaccharide can be formed by a condensation reaction between two glucose molecules.</li>
  <li>Carbohydrates can be used as an energy source for cells.</li>
  <li>Carbohydrates can be used as a structural molecule.</li>
</ol>', "ans": "a,b,c, and d", "ansSel": ["a,b", "c,d", "b,c"], "ansInfo": "" }] }
```

[...]

PR #486, opened 2022-03-08
fix HTML embedded in quizzes

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/473

#### What's this PR do?
This PR introduces some extra code to the `multiple_choice_questions_widget` turndown rule that adds support for parsing arbitrary HTML inserted into the question and answer data.  Previously, if there was HTML in the quiz data with double quotes in it, it would cause the turndown rule to throw a hard error because a call to `JSON.parse` would fail.  This happened because the data was not valid JSON.  This PR corrects that by finding the HTML, transforming it to markdown and correcting any places where single quotes are used to encapsulate the HTML.

 - Read the readme if you've never used `ocw-to-hugo` before and make sure you are set up to download courses from S3
 - Place the following in `private/courses.json`:
```
{
  "courses": [
      "7-01sc-fundamentals-of-biology-fall-2011"
  ]
}
```
 - Run `node . -i private/input -o private/test --rm --courses private/courses.json --download`
 - Inspect the output and / or spin up the course using `ocw-hugo-themes` (follow directions in the readme if you've never used it before) and ensure that the pages with quizzes on them render properly

image

### DIFF
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -21,7 +21,7 @@ class TurndownService extends OriginalTurndownService {
    * https://github.com/mixmark-io/turndown#escaping-markdown-characters
    *
    * That's good. But it does NOT (usually) escape angle brackets in the output,
-   * e.g. "&lta dog&gt"" turns into "<a dog>", which can be erroneously
+   * e.g. "&lt;a dog&gt;"" turns into "<a dog>", which can be erroneously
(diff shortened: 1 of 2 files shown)


abeglova (approved):
👍

PR #457, opened 2022-02-09
new fields for last import

gumaerc:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-to-hugo/issues/456

#### What's this PR do?
This PR brings in several new fields that were added to `ocw-data-parser` in https://github.com/mitodl/ocw-data-parser/pull/165.  It also brings in some fields that we weren't bringing in from instructors to preserve the data for use in future migrations.

#### How should this be manually tested?
 - Read the readme if you've never used `ocw-to-hugo` before and ensure you are set up for S3 access
 - Put the following in `private/courses.json`:
```
{
    "courses": [
        "22-033-nuclear-systems-design-project-fall-2011",
        "18-06-linear-algebra-spring-2010",
        "20-219-becoming-the-next-bill-nye-writing-and-hosting-the-educational-show-january-iap-2015"
    ]
}
```
 - Run `node . -i private/input -o private/output -c private/courses.json --download --rm`
 - Verify that the metadata properties listed in the attached issue are present in the converted course data, with course level metadata properties in `data/course_legacy.json` and video data in video resources under `content/resources` in each course.  Cross reference the values with the parsed json downloaded into `private/input`.

### NOTE
At the moment I am setting this PR as blocked because although https://github.com/mitodl/open-discussions/pull/3502 has merged, a full `backpopulate_ocw_data` will need to be run in `open-discussions` before the parsed JSON in the `open-learning-course-data` bucket will contain the new properties.

PR #491, opened 2022-03-16
Added highlights_text and relevant_content fields in course.json

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing

Part of issue https://github.com/mitodl/ocw-hugo-projects/issues/141

#### What's this PR do?
It adds relevant fields to course json

Verify the following fields
- related_content
- highlights_text
are parsed into the course.json

### DIFF
--- a/src/data_template_generators.js
+++ b/src/data_template_generators.js
@@ -48,7 +48,9 @@ const generateDataTemplate = (courseData, pathLookup) => ({
   open_learning_library_versions: helpers.getOpenLearningLibraryVersions(
     courseData["open_learning_library_related"]
   ),
-  legacy_uid: helpers.addDashesToUid(courseData["uid"])
+  legacy_uid: helpers.addDashesToUid(courseData["uid"]),
+  highlights_text: courseData["highlights_text"] || "",
+  related_content: courseData["related_content"] || ""
 })
 
 const generateLegacyDataTemplate = (courseData, pathLookup) => {
--- a/src/data_template_generators.js
+++ b/src/data_template_generators.js
@@ -48,7 +48,7 @@ const generateDataTemplate = (courseData, pathLookup) => ({
   open_learning_library_versions: helpers.getOpenLearningLibraryVersions(
     courseData["open_learning_library_related"]
   ),
-  legacy_uid: helpers.addDashesToUid(courseData["uid"]),
+  legacy_uid:      helpers.addDashesToUid(courseData["uid"]),
   highlights_text: courseData["highlights_text"] || "",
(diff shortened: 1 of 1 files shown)


pdpinch:
and a nit: can you change the PR title and the commit message to say something more about what fields are added?

Wassaf-Shahzad:
FYI,
The following fields are added in this PR/
- highlights_text
- relevant_content
So that legacy courses with these fields could be imported to ocw-studio

src/data_template_generators.js
```diff
@@ -48,7 +48,9 @@ const generateDataTemplate = (courseData, pathLookup) => ({
   open_learning_library_versions: helpers.getOpenLearningLibraryVersions(
     courseData["open_learning_library_related"]
   ),
-  legacy_uid: helpers.addDashesToUid(courseData["uid"])
+  legacy_uid:      helpers.addDashesToUid(courseData["uid"]),
+  highlights_text: courseData["highlights_text"] || "",
```
gumaerc, line 52: Can we also make these changes to `generateLegacyDataTemplate`?  For some history on this:

Initially the `import_ocw_course_sites` management command in `ocw-studio` parsed the same `course.json` data template that's used by `ocw-hugo-themes` to actually render courses, but we had to change this Once we started handling thils like instructors by UUID reference.  Since `ocw-to-hugo` is the original source of truth for courses in studio, the full instructor data needed to be written out for the purposes of import, but we wanted the `course.json` files to match the schema that `ocw-studio` uses exactly so we didn't have to do any special casing in the theme.  Therefore, we ended up with two versions of the course data template, one just called `course.json` and another called `course_legacy.json`.  The `course_legacy.json` file should mirror `course.json`, except for data like instructors and the course images all of the properties are included, not just the UUID.  I hope this makes sense, p/ease ask if you need clarification on anything.

gumaerc (changes requested):
This looks good, but we need to make the change in both data templates:

src/data_template_generators.js
Wassaf-Shahzad: @gumaerc 
I believe you already added these fields in the following PR
https://github.com/mitodl/ocw-to-hugo/pull/457/files#diff-19c6012e03f2a690ce54a5eb8e83d6fd6635c8674f93f08fbe7da54e573f05bc

gumaerc (approved):
:+1:

src/data_template_generators.js
Wassaf-Shahzad: No problem.

Issue #493, opened 2022-03-21
Pages in 16.90, 15.071, 6.042J don't convert to markdown


pdpinch:
### Steps to Reproduce

1. Run ocw-to-hugo on one or more of these courses: 
```
{
    "courses": [
        "6-042j-mathematics-for-computer-science-spring-2015",
        "16-90-computational-methods-in-aerospace-engineering-spring-2014",
        "15-071-the-analytics-edge-spring-2017"
    ]
}
```
2. In the output, examine pages like 

- `6-042j-mathematics-for-computer-science-spring-2015/content/pages/proofs/tp2-2/vertical-a28e46f96fa1.md`
- `6-042j-mathematics-for-computer-science-spring-2015/content/pages/proofs/tp4-3/set-theory-axioms-optional-0.md`
- `16-90-computational-methods-in-aerospace-engineering-spring-2014/content/pages/numerical-integration-of-ordinary-differential-equations/discretizing-odes/_index.md`
- `15-071-the-analytics-edge-spring-2017/content/pages/logistic-regression/modeling-the-expert-an-introduction-to-logistic-regression/quick-question-152.md`
- `14-01sc-principles-of-microeconomics-fall-2011/content/pages/unit-6-topics-in-intermediate-microeconomics/international-trade.md`

### Expected Behavior

- The pages should contain head matter, followed by markdown 
- If the HTML to markdown converter fails, it should log an error so we can debug and investigate

### Actual Behavior

- instead of markdown, the pages contain HTML

pdpinch:
a slight larger list of courses with similar issues appears in https://github.com/mitodl/ocw-studio/issues/1119

pdpinch:
Another course/page that might have the same issue? https://ocwnext.odl.mit.edu/courses/20-011j-statistical-thermodynamics-of-biomolecular-systems-be-011j-spring-2004/pages/simulations/

mbertrand:
These are all pages with quizzes that aren't covered by existing turndown rules in ocw-to-hugo

PR #497, opened 2022-03-22
fix overzealous turndown filter on quiz rule

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/493

#### What's this PR do?
In https://github.com/mitodl/ocw-to-hugo/pull/418 we added support for parsing multiple choice type quizzes embedded in legacy OCW courses.  One of the turndown rules that does this, `multiple_choice_question`, was filtering on:

```
        node.getAttribute("class") === "problem_question" &&
        node.getElementsByClassName("problem_radio_input")
```

The issue with this is that even if `getElementsByClassName` doesn't find anything, it still returns an object.  This means that HTML that shouldn't have been matching that rule was being matched and processed as if it were an multiple choice question.  I corrected this by changing it to:

```
        node.getAttribute("class") === "problem_question" &&
        node.getElementsByClassName("problem_radio_input").length > 0
```

A try / catch statement with Winston file logging was also added to the `html2markdown` function so any errors converting html to markdown are logged.

 - Read the readme if you've never used `ocw-to-hugo` before and make sure you're set up to download courses from S3
 - Place the following in `private/courses.json`:
```
{
    "courses": [
        "6-042j-mathematics-for-computer-science-spring-2015",
        "16-90-computational-methods-in-aerospace-engineering-spring-2014",
        "15-071-the-analytics-edge-spring-2017",
        "14-01sc-principles-of-microeconomics-fall-2011"
    ]
}
```
 - Run `node . -i private/input -o private/output -c private/courses.json --download --rm`
 - Inspect the following pages and make sure there is no raw HTML:
```
 - 6-042j-mathematics-for-computer-science-spring-2015/content/pages/proofs/tp2-2/vertical-a28e46f96fa1.md
 - 6-042j-mathematics-for-computer-science-spring-2015/content/pages/proofs/tp4-3/set-theory-axioms-optional-0.md
 - 16-90-computational-methods-in-aerospace-engineering-spring-2014/content/pages/numerical-integration-of-ordinary-differential-equations/discretizing-odes/_index.md
 - 15-071-the-analytics-edge-spring-2017/content/pages/logistic-regression/modeling-the-expert-an-introduction-to-logistic-regression/quick-question-152.md
 - 14-01sc-principles-of-microeconomics-fall-2011/content/pages/unit-6-topics-in-intermediate-microeconomics/international-trade.md
```

### DIFF
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -488,7 +488,7 @@ turndownService.addRule("multiple_choice_question", {
     if (node.nodeName === "DIV" && node.getAttribute("class")) {
       if (
         node.getAttribute("class") === "problem_question" &&
-        node.getElementsByClassName("problem_radio_input")
+        node.getElementsByClassName("problem_radio_input").length > 0
       ) {
         return true
       }
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -504,7 +504,7 @@ turndownService.addRule("multiple_choice_question", {
         if (
           child.nodeName === "DIV" &&
           child.getAttribute("class").includes("choice") &&
-          child.getElementsByClassName("problem_radio_input") > 0
+          child.getElementsByClassName("problem_radio_input").length > 0
         ) {
           return true
         }
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -750,8 +750,7 @@ function html2markdown(text) {
   let markdown = ""
   try {
     markdown = turndownService.turndown(text)
-  }
-  catch (err) {
+  } catch (err) {
     loggers.fileLogger.error(`Error converting html to markdown: ${err}`)
   }
   return markdown
--- a/src/turndown.js
+++ b/src/turndown.js
@@ -504,7 +504,7 @@ turndownService.addRule("multiple_choice_question", {
         if (
           child.nodeName === "DIV" &&
           child.getAttribute("class").includes("choice") &&
-          child.getElementsByClassName("problem_radio_input")
+          child.getElementsByClassName("problem_radio_input") > 0
         ) {
           return true
         }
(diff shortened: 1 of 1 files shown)


mbertrand (changes requested):
The markdown is generated for these pages but the bulk of the quiz content is missing.  For example, `6-042j-mathematics-for-computer-science-spring-2015/content/pages/proofs/tp2-2/vertical-a28e46f96fa1.md` is missing the select options and answers shown here:  https://ocw.mit.edu/courses/electrical-engineering-and-computer-science/6-042j-mathematics-for-computer-science-spring-2015/proofs/tp2-2/vertical-a28e46f96fa1/

Shouldn't this quiz content be converted to markdown, as is done with some other quiz formats?

src/turndown.js
```diff
@@ -488,7 +488,7 @@ turndownService.addRule("multiple_choice_question", {
     if (node.nodeName === "DIV" && node.getAttribute("class")) {
       if (
         node.getAttribute("class") === "problem_question" &&
-        node.getElementsByClassName("problem_radio_input")
+        node.getElementsByClassName("problem_radio_input").length > 0
```
mbertrand, line 491: There's a `child.getElementsByClassName("problem_radio_input")` in the `replacement:` section below, maybe add `.length > 0` there too just in case?

mbertrand (approved):
Approving since the missing quiz content will be dealt with later in a separate PR.  Sorry for giving the extra work to you, I had gotten this far with the issue yesterday but assumed that extra turndown rules had to be added to correctly render the quiz content on the affected pages.

src/turndown.js
gumaerc: Nothing with no results on that class search should get through the filter, but I added this anyway because it's a sensible precaution.

Issue #512, opened 2022-05-31
page descriptions missing


gumaerc:
In `ocw-to-hugo`, we are pulling in the description property of the course home page as the course level description and resource descriptions are being pulled in, but not pages.  The `description` property of `course_page` items should be pulled in and set in the front matter of pages.

PR #513, opened 2022-05-31
pull in description property of course pages

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-to-hugo/issues/512

#### What's this PR do?
This PR pulls in the `description` property of `course_pages` items and sets it on the front matter of pages.  These are plain text descriptions that will be used for `meta` tags in the `head` for the purpose of search indexing.

 - Read the readme if you've never used `ocw-to-hugo` before and make sure you are set up to download courses from S3
 - Run `node . -i private/input -o private/output -c course_json_examples/example_courses.json --download --rm`
 - Inspect the courses output to `private/output` and ensure that the pages under `content/pages` have a `description` property, it is plain text and accurate to the page it belongs to

### DIFF
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -165,7 +165,7 @@ const generateCourseSectionFrontMatter = (
   const courseSectionFrontMatter = {
     uid:         helpers.addDashesToUid(pageId),
     title:       helpers.replaceIrregularWhitespace(title),
-    description: description
+    description: helpers.replaceIrregularWhitespace(description)
   }
 
   if (parentUid) {
--- a/src/markdown_generators.js
+++ b/src/markdown_generators.js
@@ -126,7 +126,8 @@ const generateMarkdownRecursive = (page, courseData, pathLookup) => {
     courseData["short_url"],
     page["is_media_gallery"],
     helpers.getVideoUidsFromPage(page, courseData),
-    page["type"]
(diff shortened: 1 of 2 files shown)


src/markdown_generators.js
```diff
@@ -155,14 +156,16 @@ const generateCourseSectionFrontMatter = (
...
 ) => {
   /**
     Generate the front matter metadata for a course section
     */
   const courseSectionFrontMatter = {
-    uid:   helpers.addDashesToUid(pageId),
-    title: helpers.replaceIrregularWhitespace(title)
+    uid:         helpers.addDashesToUid(pageId),
+    title:       helpers.replaceIrregularWhitespace(title),
+    description: description
```
mbertrand, line 168: Do we need to worry about any potential "irregular whitespace" here like in the title?

mbertrand (approved):
LGTM, I had one question but assuming answer is no then 👍

src/markdown_generators.js
gumaerc: Good call.  I ran this through all courses both with and without and there were a lot of descriptions that ended up with literal line breaks (`\r\n`) written into them.  Passing it through `replaceIrregularWhitespace` fixed that.

